PR #1, opened 2022-09-21
Improved DB performance & loading times

chiragnotionforms:
When there was more forms, was took more time
So Improved DB performance & loading times


JhumanJ (approved):
Great!

PR #14, opened 2022-10-17
Handle dirty form editor

chiragnotionforms:
On Form edit if user edit something and try to leave page then without saving, then display confirmation alert


chiragnotionforms:
Will do this in new PR 
so no need this  old PR

PR #20, opened 2022-10-31
9a3d4 form page

chiragnotionforms:
Hi Julien, 
I did form page from my side, can you please do some polishing
Here are some known points

For below modals can you please change icons
- Generate new form link
- Url Form Prefill
- Create template

image
When there is big block Copy button height is big it should not increase height

Once we move to another tabs and again come back to submissions tab then something infinite there so browser stop working, can you please check it

Thanks for your help :)


JhumanJ (changes requested):
Hey Chirag, thank you for this. The design overall is pretty well respected so thank you! I polished few things and fixed others. About the icons in the modal, here's a general rule: if the button triggering the modal has an icon, then just use the same icon in the modal. 

Now here are some changes to do please:
- The code is way too long and too complicated. We're open-sourcing this so we should do the best we can :) I'm mostly talking about the `show.vue` file that is 500 lines long! We need to split it in different logical components so that the code for this page is much easier to read. On the image below, I've drawn for you a list of all components that should be created (each red square should be it's own component). Let me know if you have any questions about this. You can put all of these new components in a new folder: `/resources/js/components/pages/forms/show/`.
- We want to split the 3 different tabs in 3 different pages (using vue-router). By doing so, each tab will have it's own URL (`.../submissions`, `.../analytics`, `.../share`) which will be cleaner. To do this, you can look at how it works for the [settings routes](https://github.com/JhumanJ/OpnForm/blob/eaf6e9b212f5964a4d76257579eea6791cbabb9b/resources/js/router/routes.js#L18-L31). Essentially, we have a main show component that has `router-view` where the content should be replaced depending on the route. Each tab button (blue rectangle on the image) is then replaced by a router-link. Please checkout the [`settings/index.vue` file](https://github.com/JhumanJ/OpnForm/blob/eaf6e9b212f5964a4d76257579eea6791cbabb9b/resources/js/pages/settings/index.vue#L76) to see how it's implemented. and feel free to reach out if you have any questions.

Thank you!

image

resources/js/pages/forms/show/index.vue
```diff
@@ -0,0 +1,226 @@
...
+        {
+          name: 'Analytics',
+          route: 'forms.show.analytics'
+        },
+        {
+          name: 'Share',
+          route: 'forms.show.share'
+        }
+      ],
+      currentTab: 1,
```
JhumanJ, line 153: No need for this anymore

resources/js/pages/forms/show/analytics.vue
```diff
@@ -0,0 +1,34 @@
...
+import FormStats from '../../../components/open/forms/components/FormStats'
+
+export default {
+  components: {FormStats},
+  props: {
+    form: { type: Object, required: true }
+  },
+
+  metaInfo() {
+    return {title: 'Form Analytics'}
```
JhumanJ, line 20: Can you include the form name in the title (if present)

resources/js/pages/forms/show/share.vue
```diff
@@ -0,0 +1,47 @@
...
+    EmbedCode,
+    UrlFormPrefill,
+    RegenerateFormLink
+  },
+  props: {
+    form: { type: Object, required: true }
+  },
+
+  metaInfo() {
+    return {title: 'Form Share'}
```
JhumanJ, line 33: Same for form name (if form defined)

resources/js/pages/forms/show/submissions.vue
```diff
@@ -0,0 +1,31 @@
...
+import FormSubmissions from '../../../components/open/forms/components/FormSubmissions'
+
+export default {
+  components: {FormSubmissions},
+  props: {
+    form: { type: Object, required: true }
+  },
+
+  metaInfo() {
+    return {title: 'Form Submissions'}
```
JhumanJ, line 17: Same here

JhumanJ (changes requested):

PR #46, opened 2022-12-22
Some fixes and installation documentation

MarceauKa:
Hi guys! I was here to write some doc wheb I have issues trying to install OpnForm locally.

So, you'll find:
- Documentation for the installation
- Some fixes in a migration: useless DB Transaction (not so useless after reading the process, but don't work with MariaDB here), and fixes malformed foreign keys constraints
- .gitignore for the `storage/clockwork` folder

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,45 @@ It takes 1 minute to try out the builder for free. You'll have high availability
 
 ## Self-hosting
 
-This section hasn't been written yet, we need to work on this. 
+🚧 This section is under construction!
+
+### Requirements
+
+- PHP >= 8.0
+- MySQL/MariaDB or PostgreSQL
+- Node.js and NPM/Yarn/... to compile assets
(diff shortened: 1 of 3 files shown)


JhumanJ (changes requested):
Hey Marceau! 
Merci a lot for your help 🙏

Just a few thoughts about the install instruction:
- They're great but they're very "Laravel specific". So people outside of the Laravel world would be able to follow them. Maybe this section should be renamed something like Local Installation for Laravel devs? Ideally, I'd like a generic docker setup to be the main install process
- It seems like you're using valet in your instructions, it might be able to add few instructions about it (`valet park`, `valet secure` etc.)
Thank you again!

Et bonne fetes 🎅

JhumanJ (approved):
Overall these instructions are better than nothing, thanks for your contribution!

PR #58, opened 2023-01-03
Update Form.php

brianmdavis:
getViewsCountAttribute is using a raw SQL statement that is specific to pgsql, it fails on MySQL/MariaDB.  Added a conditional to use mysql specific code to access data in the json array.  Crude, but no more so than the original?


JhumanJ:
Thanks, we're one step closer to supporting Mysql :)

PR #64, opened 2023-01-06
Hide form title via URL param

chiragnotionforms:
Now user can hide form title via pass param `hide_title=true`  in URL


chiragnotionforms:
@JhumanJ 
close this PR because another PR created for this feature

Issue #76, opened 2023-01-25
Bug: Help button in navbar/help.opnform.com not working


jeffrey-zang:
![image](https://user-images.githubusercontent.com/66485719/214596131-ce32af52-11a3-403f-90f9-47d18a89bee0.png)
This link doesn't do anything
When I went to help.opnform.com, it looked like this
![image](https://user-images.githubusercontent.com/66485719/214596260-7ba449fd-6fa5-4c07-bcb4-50be4b4a2b95.png)
Looks like someone needs to renew the Crisp plan
In the meantime, we could redirect to the [discussions](https://github.com/JhumanJ/OpnForm/discussions) page

JhumanJ:
Hey, you're right we just stopped paying for Crisp as it was too expensive for a free project. Good idea, we're going to redirect to the Github discussion pages!
Thanks!

jeffrey-zang:
@JhumanJ can i make a pr pls?

PR #77, opened 2023-01-25
fix: navbar help link

jeffrey-zang:
#76


jeffrey-zang:
btw what are you guys deploying the website with? just curious

JhumanJ (changes requested):
Thanks for this! It wasn't the case here but we try to use `links.php` config file to centralize all links. You can then use `window.config.links.help_url` in the javascript to consume it. Could you please make the change? 🙏

btw, we use https://vapor.laravel.com/ for deployments

jeffrey-zang:
@JhumanJ not sure if I did this right, does it look ok?

JhumanJ (approved):
Great, thank you!

PR #131, opened 2023-05-21
Translate fields in different languages

terryMarsault:
Add the possibility to translate forms in different languages. For exemple, 'From' and 'To' labels in a date field are by default written in english. I would like to translate them in french but I can't. The same for the 'Upload files' field which contains english labels. etc.

image

image

__

**In my opinion, no need of dictionaries, language files and stuff like that. I think we should just be able to write whatever we want directly in the configuration of the fields in form editor, it would be fairly enough.**


JhumanJ:
Hey @terryMarsault, I'm confused, there are 49 files updated in this PR, and you seem to be merging one of our branch? Can you please explain?

terryMarsault:
@JhumanJ, sorry, I'm a beginner at GitHub (which looks like a real gaz factory for me...), I just want to pull a request without forking or editing the code

JhumanJ (changes requested):
Hey @terryMarsault, je ne vais pas merge ça du coup, c'est compliqué. Si tu veux échanger et qu'on puisse trouver des manières de collaborer, n'hésite pas à m'envoyer un email à julien@nahum.net. Merci pour ton intéret au projet!

PR #134, opened 2023-06-08
Norwegian translation

wef0rm:
I've translated the template-files that I found.

Are there any for the web-interface itself, or is that hardcoded?


JhumanJ (approved):
Thank for your effort. I'll merge this, but it won't have an impact on the project as of today, as we do not support internationalization. All of the front-end files have hard-coded translation strings, so for now, we just can't change them.

PR #141, opened 2023-07-03
Added German translation

f-str:
I have added translations for the German language at all places I managed to find.


JhumanJ (approved):
Thank for your effort. I'll merge this, but it won't have an impact on the project as of today, as we do not support internationalization. All of the front-end files have hard-coded translation strings, so for now, we just can't change them.

PR #173, opened 2023-08-19
fix: support both mysql + postgre

### LINKED ISSUE #171: MySQL support
This project was developed with PostgreSQL. Many people would like to make it compatible with MySQL. This task is about making the required changes so that it can also work with MySQL.

[See this previous issue about MySQL support](https://github.com/JhumanJ/OpnForm/issues/52)

rishi-raj-jain:
fix #171 
/claim #171


rishi-raj-jain:
The mysql one now works just fine, how do I test postgre locally?

rishi-raj-jain:
Do you want me to convert this to draft or me asking questions without converting it is fine and I can continue committing if needed?

rishi-raj-jain:
The mysql one now works just fine, how do I test postgre locally?

rishi-raj-jain:
The mysql one now works just fine, how do I test postgre locally?

rishi-raj-jain:
@JhumanJ Did you get a chance at it?

JhumanJ:
Hey @rishi-raj-jain thank you for this! I'm currently travelling, but I'll review this in the next 2 days!

rishi-raj-jain:
@JhumanJ Did you get a chance at it?

database/migrations/2022_08_18_133641_add_removed_properties_to_forms.php
```diff
@@ -14,7 +14,7 @@
     public function up()
     {
         Schema::table('forms', function (Blueprint $table) {
-            $table->jsonb('removed_properties')->default('[]');
+            $table->text('removed_properties');
```
JhumanJ, line 17: Removing defaults breaks the application (see the tests in the workflow run). If you can't have them in the database layer, you at least need them at the application layer (Maybe in `$attributes` on the model?)

database/migrations/2022_09_22_092205_create_templates_table.php
```diff
@@ -20,7 +20,7 @@ public function up()
             $table->string('slug');
             $table->text('description');
             $table->string('image_url');
-            $table->jsonb('structure')->default('{}');
```
JhumanJ, line 23: Same for defaults

database/migrations/2022_08_18_133641_add_removed_properties_to_forms.php
JhumanJ: Why not json filed also? `text` is not the right type here

JhumanJ (changes requested):
Hey, thank you for this. Tests won't run because of broken default values. Also please don't replace `json` with `text`

rishi-raj-jain:
@JhumanJ 

> Tests won't run because of broken default values. Also please don't replace json with text

How can I run them locally? Want to test it out before making the next commit.

JhumanJ:
> @JhumanJ
> 
> > Tests won't run because of broken default values. Also please don't replace json with text
> 
> How can I run them locally? Want to test it out before making the next commit.

`php artisan test`

JhumanJ:
Closing this for now. Please re-open if needed.

CISPDEVELOP:
Hello, @JhumanJ  I know that this PR was closed but I want to suggest to continue support for mysql and future migrations are compatible across different engines.
. Because in #176  the migration 2023_08_23_100710_add_notification_settings_to_forms.php
keep creating columns that are not supported in mysql. 
I don't know if this is contemplated in the project or should be manually modified

thanks

rishi-raj-jain:
Hey hey, @JhumanJ how can I re-open this?

rishi-raj-jain:
I don't see any option of the same

CISPDEVELOP:
MIGRATION FAILS

rishi-raj-jain:
@JhumanJ can you re-open this!?

CISPDEVELOP:
@rishi-raj-jain 
The main problem is wanting to store default values in the DB, since some engines do not support it for certain types of data, and it would also consume storage unnecessarily.
as Julien said, Making these columns null implies that the model will have to resolve when the column is empty and return the expected/valid default value, but it would make it compatible at least.
A simple example for text columns, imagine storing 500 forms and all with default values in 3 or more text type columns, I think it would be a storage and translation problem in the future that the model can solve.

Please consider that part, thanks for the support!

JhumanJ:
Hi there. I'm happy to be re-opening this, but I think we need to agree on an approach. We can't have default at the database-level when using mysql, so I think we need to move all default values to the application level, more precisely on the models. The best way of doing this is probably to move defaults to models' attributes:
```php
protected $attributes = [
        'status' => self::STATUS_UNCONFIRMED,
        'role_id' => self::ROLE_PUBLISHER,
];
```
I'm also curious about the performances of mysql `json` versus postgreSQL's `jsonb`, but I guess this is less important for now.

rishi-raj-jain:
> We can't have default at the database-level when using mysql, so I think we need to move all default values to the application level

Yeah that'd take care of both mysql & postgres

database/migrations/2022_08_18_133641_add_removed_properties_to_forms.php
rishi-raj-jain: I don't think that json or jsonb work in mysql. Hence, we need to parse it as text

rishi-raj-jain:
@JhumanJ 

> Also please don't replace json with text

I don't think that mysql in any form supports JSON, and the right type shall be text for mysql I believe?

rishi-raj-jain:
Ah okay, I see that mysql has a json type, changing with that

rishi-raj-jain:
Should probably add https://github.com/imagemin/imagemin/issues/168#issuecomment-314313374 to the documentation as it is required to run the project on macOS

database/migrations/2022_08_18_133641_add_removed_properties_to_forms.php
rishi-raj-jain: resolved

rishi-raj-jain:
also https://getgrav.org/blog/macos-ventura-apache-multiple-php-versions

rishi-raj-jain:
Screenshot 2023-09-07 at 4 14 24 PM

While testing it, this is what I see, but can't seem to find the innodb in the code, how to fix this with `php artisan test`?

rishi-raj-jain:
@JhumanJ Did you get a chance at my error posted above?

CISPDEVELOP:
> @JhumanJ Did you get a chance at my error posted above?
Hi @rishi-raj-jain 
it seems strange to me, This seems to be a problem of database permissions rather than the definition of migration.
Try something like: 
GRANT SELECT ON mysql.innodb_index_stats TO 'your_user'@'your_host';
FLUSH PRIVILEGES;

I hope it helps you

JhumanJ:
I'm not familiar with mySQL I'm afraid I can't help

JhumanJ:
Closing for now.

Issue #170, opened 2023-08-18
Phone Input Component


JhumanJ:
Currently the input component used for the `phone` type is a basic `<text-input>`. We want to have a dedicated component for the phone type. Ex:
![image](https://github.com/JhumanJ/OpnForm/assets/11312432/321fac0e-c78d-4593-b644-14a99c9c2ff7)

The goal of this issue is to create this `PhoneInput.vue` component, and to use it for the `phone` type. Required data for the countries can be placed as a json in `resources/data`.

**Requirements**
- It should work with all countries and contain nice flags
- It must work will existing form themes (see `form-theme.js`). Themes can be changed in the customization section of the form editor
- The component must be lazy-loaded to avoid loading all of its data on every page. Ex here: [/resources/js/components/forms/index.js#L41](https://github.com/JhumanJ/OpnForm/blob/96ef9d7c25fade952740eff47c4d469be07b810b/resources/js/components/forms/index.js#L41)
- The dropdown must be using our `VSelect.vue` component
- It should handle clashes for instance, for French number you don't need `+33` and start with a `0`, so if user selects `+33` for France, and enters `0123456789`, the resulting value should not be `+330123456789` but `+330123456789`
- It should have a dedicated phone validation rule in the back-end in `AnswerFormRequest`. The number must start with a valid country code and have the right length

Please feel free to reach out if you have any question!

Ashutosh-Bhadauriya:
@JhumanJ I would like to take this 🙂
/attempt #170

rishi-raj-jain:
I'm next in line!

sagar-kashyap:
> the resulting value should not be `+330123456789` but `+330123456789`

@JhumanJ can you clarify your 2nd last point in requirement.

JhumanJ:
Hey guys. So essentially remove the leading 0 if there is one before pre-pending the country code

Ashutosh-Bhadauriya:
Hey @JhumanJ I'm getting this error, when running npm run dev, any idea. I tried googling found some fixes but it didn't help. Also tried adding this to vite.config.js, but no success
`  server: {
    https: true,
    host: 'localhost'
  },`

![image](https://github.com/JhumanJ/OpnForm/assets/62984427/10ca3ceb-aa72-4665-9fb9-e660792ed3ca)

JhumanJ:
Hey @Ashutosh-Bhadauriya, I've just pushed something to fix this, please pull an try again.

[...]

PR #188, opened 2023-09-04
Feat/170 phone input structure

Ahtritus:
ENH:  #170 - Added phone input component with backend validation rules
image

### DIFF
--- a/app/Http/Requests/AnswerFormRequest.php
+++ b/app/Http/Requests/AnswerFormRequest.php
@@ -150,7 +150,6 @@ class AnswerFormRequest extends FormRequest
     {
         switch ($property['type']) {
             case 'text':
-            case 'phone_number':
             case 'signature':
                 return ['string'];
             case 'number':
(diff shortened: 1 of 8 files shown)


JhumanJ (changes requested):
Hi, thanks for the great work. I've just gave it a try, but unfortunately this doesn't work for us, as the created input doesn't match the style of the inputs we already have (regardless of the theme). That's why I didn't recommend going with a library.

Ahtritus:
> Hi, thanks for the great work. I've just gave it a try, but unfortunately this doesn't work for us, as the created input doesn't match the style of the inputs we already have (regardless of the theme). That's why I didn't recommend going with a library.

Hey, I made a couple of changes to make sure it works with all the themes but I'm guessing you're talking about the dropdown select element?

image


If that's the case, I'll perhaps give it some more time to create a custom input, only problem there is the flags which are a bit tricky.

JhumanJ:
There's not only the dropdown, but also the colored shadow when input is focused and some other inconsistencies unfortunately. As mentioned, I think the flag issue can easily be resolved with a small library that does just this :)

Ahtritus:
Makes sense. I did go ahead and created a custom dropdown phone input component. Closing this and raising a new PR for that

PR #213, opened 2023-09-29
Phone input prefill country issue fixed

formsdev:


app/Http/Requests/AnswerFormRequest.php
```diff
@@ -246,6 +246,14 @@ protected function prepareForValidation()
             }
         });
 
+        $countryCodeMapper = json_decode(file_get_contents(resource_path('data/country_code_mapper.json')), true);
```
JhumanJ, line 249: We already did `collect(properties)` to improve speed we should avoid doing this many times. You can run the `->each` right after the `->filter()`

JhumanJ (changes requested):
It's great, just a small optimization when preparing for validation

JhumanJ (approved):

PR #228, opened 2023-10-23
feat(s3-aws): Implement use_path_style_endpoint for Minio usage 

valentin2105:
Hello

I tryed to use the S3 Storage with Minio (open-source S3 platform), but the `aws path style` is problematic.

It would be nice if we can tweak that from ENV VAR.

Thank you

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -63,6 +63,7 @@ return [
             'bucket' => env('AWS_BUCKET'),
             'url' => env('AWS_URL'),
             'endpoint' => env('AWS_ENDPOINT'),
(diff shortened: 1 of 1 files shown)


JhumanJ (changes requested):
Thanks for the PR. That works for me, but can you make it true by default to keep the preserve behavior and avoid breaking existing deployments?

valentin2105:
@JhumanJ the change is done.

JhumanJ (approved):

JhumanJ (approved):
Great thanks!

JhumanJ:
Merged, thank you!

PR #242, opened 2023-11-15
fix: migrations not passing on pgsql

### LINKED ISSUE #241: MySQL Feature Broke Postgres Migrate
I was setting up a new install on Postgres and received this error on a PHP migrate, however PHP migrate worked fine with a mySQL database.

  SQLSTATE[42883]: Undefined function: 7 ERROR:  function json_array() does not exist at character 1882
HINT:  No function matches the given name and argument types. You might need to add explicit type casts. (SQL: create table "forms" ("id" bigserial primary key not null, "workspace_id" bigint not null, "title" varchar(255) not null, "slug" varchar(255) not null, "properties" json not null, "created_at" timestamp(0) without time zone null, "updated_at" timestamp(0) without time zone null, "notifies" boolean not null default '0', "description" text null, "submit_button_text" text not null default ('Submit'), "re_fillable" boolean not null default '0', "re_fill_button_text" text not null default ('Fill Again'), "color" varchar(255) not null default '#3B82F6', "uppercase_labels" boolean not null default '1', "no_branding" boolean not null default '0', "hide_title" boolean not null default '0', "submitted_text" text not null default ('Amazing, we saved your answers. Thank you for your time and have a great day!'), "dark_mode" varchar(255) not null default 'auto', "webhook_url" varchar(255) null, "send_submission_confirmation" boolean not null default '0', "logo_picture" varchar(255) null, "cover_picture" varchar(255) null, "redirect_url" varchar(255) null, "custom_code" text null, "notification_emails" text null, "theme" varchar(255) not null default 'default', "database_fields_update" jsonb null, "width" varchar(255) not null default 'centered', "transparent_background" boolean not null default '0', "closes_at" timestamp(0) without time zone null, "closed_text" text null, "notification_subject" varchar(255) not null default 'We saved your answers', "notification_body" text not null default ('<p>Hello there 👋 <br>This is a confirmation that your submission was successfully saved.</p>'), "notifications_include_submission" boolean not null default '1', "use_captcha" boolean not null default '0', "can_be_indexed" boolean not null default '1', "password" varchar(255) null, "notification_sender" varchar(255) not null default 'OpenForm', "tags" jsonb not null default (JSON_ARRAY())))

nhedger:
Not sure why, but it appears that the recently introduced MySQL has broken the PostgreSQL migrations. Since all tests passed, I was not expecting that, but here's a fix.
 
The affected migrations now run alternative code depending on the driver being used.

Fixes #241 

---

I'll dig more to see why the tests did not catch this.

### DIFF
--- a/.github/workflows/laravel.yml
+++ b/.github/workflows/laravel.yml
@@ -12,7 +12,7 @@ jobs:
     services:
       postgres:
         # Docker Hub image
-        image: postgres
+        image: postgres:14
         # Provide the password for postgres
         env:
           POSTGRES_PASSWORD: postgres
--- a/database/migrations/2021_05_19_140326_create_forms_table.php
+++ b/database/migrations/2021_05_19_140326_create_forms_table.php
@@ -3,6 +3,7 @@
 use Illuminate\Database\Migrations\Migration;
 use Illuminate\Database\Query\Expression;
 use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Schema;
 
 class CreateFormsTable extends Migration
@@ -14,7 +15,9 @@ class CreateFormsTable extends Migration
      */
     public function up()
     {
-        Schema::create('forms', function (Blueprint $table) {
+        $driver = DB::getDriverName();
+
+        Schema::create('forms', function (Blueprint $table) use ($driver) {
             $table->id();
             $table->foreignIdFor(\App\Models\Workspace::class,'workspace_id');
             $table->string('title');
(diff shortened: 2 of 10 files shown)


nhedger:
Alright, so what happened is that I was running the tests against PostgreSQL 16.x locally, and the GitHub Actions workflow does the same, and the reason it does so is because no version is specific in the image name, and so it falls back to the latest version (16.x at the time).

I'll update the workflow so that it uses a fixed version of `postgres:14` for running the tests

nhedger:
If this is merged, it might be a good idea to specify the minimum MySQL and PostgreSQL versions in the README.

JhumanJ (approved):

JhumanJ:
Thanks a lot for the followup PR @nhedger !

PR #335, opened 2024-03-06
20334 implement desktop camera upload feature

madassdev:
## Summary by CodeRabbit

- **New Features**
	- Improved camera upload feature with conditional rendering based on user settings.
	- Added a button to open the webcam for camera uploads.
	- Updated layout and styling of camera upload component for better user experience.
	- Introduced a checkbox in forms to enable camera uploads for image fields.


JhumanJ (changes requested):
Great work! Some feedback:
CleanShot 2024-03-12 at 18 39 51@2x

1. The snap button should have the same background color (and same text color) than the submit button (essentially use the form-color). The icon color should be computed based on the form color (like it's done for the submit button

2. Form Theme Styling should be respected for the camera upload (missing shadow, not rounded etc.). Feel free to add a new entry in `form-themes.js` for the `fileInput`
CleanShot 2024-03-12 at 18 41 48@2x

3. Currently the process is like this: 1. click the camera button, 2. confirm, 3 see uploaded image. I think to simplify we should get rid of step 2 and go directly from 1 to 3

CleanShot 2024-03-12 at 18 43 43@2x

4. Multiple file upload isn't working with the camera upload feature. I have the multiple file option enabled, but I can't take more pictures (although I still can upload files)
CleanShot 2024-03-12 at 18 46 17@2x

5. it seems like the picture taken has the size of the CameraUpload component size, is it possible to take larger picture? Full size ideally?

6. I think we should disable this on mobile since they have direct access to the camera. Make to keep it simple we can hide by default (with css `hidden`) and show it on larger screens (`sm:block`)?

Thank you!

JhumanJ (approved):

JhumanJ (approved):

JhumanJ (approved):

Issue #337, opened 2024-03-07
cant get app running in a subdirectory


gilesbradshaw:
I am running with docker:

```
version: "3.7"

services:
  opnform:
    image: jhumanj/opnform:latest
    volumes:
      - forms:/persist
      - ./nuxt.env:/app/client/.env
      - ./laravel.env:/app/.env
    networks:
      - externalnet

volumes:
  forms:
networks:
  externalnet:
    driver: overlay
    external: true

```

/app/.env has:

```
APP_URL=https://subdomain.mydomain.com/opnform

```

/app/client/.env has 

```
NUXT_PUBLIC_APP_URL=https://subdomain.mydomain.com/opnform/
NUXT_PUBLIC_API_BASE=https://subdomain.mydomain.com/opnform/api

```

However the system shows symptoms of these not being respected (resources trying to load from root).

I also tried putting invalid domain and it doesn't seem to make any difference.

I wonder if app just won't run in a sub directory or if it's something else.

To be clear I am using a reverse proxy to incorperate this app into my web site  so when external request request  https://subdomain.mydomain.com/opnform/_nuxt/moderator.At5Sz9-h.js/ is made you will get a request on /_nuxt/moderator.At5Sz9-h.js/ - but I would expect link on web site to be /opnform/_nuxt/moderator.At5Sz9-h.js/

gilesbradshaw:
k looking at code I can see it doesn't do this..

eg

```
 src="/img/pages/ai_form_builder/background-pattern.svg" 

```

JhumanJ:
Hi @gilesbradshaw we haven't built this to be run within a subdirectoty and we don't have any short terms plans to. I would think that a reverse proxy would work, but I'm not sure. I recommend you look into doing this for a nuxt app. Feel free to contribute the repo if you manage to set it up. Best of luck!

PR #340, opened 2024-03-09
allow app to be run in a sub folder

gilesbradshaw:
This allows opn form to be run in a subdirectory.
I wanted this so I can authenticate form submissions made on my web site.  I have a proxy which gets the user from http-ony cookies and then strips the cookies before sending the post on to opnform.

the client env is as follows:

```env
NUXT_PUBLIC_APP_URL=https://mysite.com/forms
NUXT_PUBLIC_API_BASE=https://mysite.com/forms/api
NUXT_APP_BASE_URL=/forms/
```

nb the NUXT_APP_BASE_URL value (nuxt uses this)

I struggled with getting this value in runtimeConfig.js (not sure why 🤷) so in the code I use the full NUXT_PUBLIC_APP_URL value.

discussed on issue  #337

### DIFF
--- a/app/Exceptions/VerifyEmailException.php
+++ b/app/Exceptions/VerifyEmailException.php
@@ -14,7 +14,7 @@ class VerifyEmailException extends ValidationException
     {
         return static::withMessages([
             'email' => [__('You must :linkOpen verify :linkClose your email first.', [
-                'linkOpen' => '<a href="/email/resend?email='.urlencode($user->email).'">',
+                'linkOpen' => '<a href="'.env("APP_URL").'/email/resend?email='.urlencode($user->email).'">',
                 'linkClose' => '</a>',
             ])],
         ]);
--- a/app/Providers/RouteServiceProvider.php
+++ b/app/Providers/RouteServiceProvider.php
@@ -30,7 +30,7 @@ class RouteServiceProvider extends ServiceProvider
         $this->registerGlobalRouteParamConstraints();
 
         $this->routes(function () {
-            Route::middleware('api')
+            Route::middleware($_ENV['APP_API_PREFIX'] ?? 'api')
                 ->namespace($this->namespace)
                 ->group(base_path('routes/api.php'));
         });
@@ -43,7 +43,7 @@ class RouteServiceProvider extends ServiceProvider
      */
     protected function configureRateLimiting()
     {
-        RateLimiter::for('api', function (Request $request) {
+        RateLimiter::for($_ENV['APP_API_PREFIX'] ?? 'api', function (Request $request) {
             return Limit::perMinute(60)->by(optional($request->user())->id ?: $request->ip());
         });
     }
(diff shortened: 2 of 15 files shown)


app/Exceptions/VerifyEmailException.php
```diff
@@ -14,7 +14,7 @@ public static function forUser($user)
     {
         return static::withMessages([
             'email' => [__('You must :linkOpen verify :linkClose your email first.', [
-                'linkOpen' => '<a href="/email/resend?email='.urlencode($user->email).'">',
```
JhumanJ, line 17: For this we should use the `front_url` helper function

app/Providers/RouteServiceProvider.php
```diff
@@ -30,7 +30,7 @@ public function boot()
         $this->registerGlobalRouteParamConstraints();
 
         $this->routes(function () {
-            Route::middleware('api')
```
JhumanJ, line 33: Hum we shouldn't not change the middleware name this should be left as it is - this is a middleware name, not a URL prefix

app/Providers/RouteServiceProvider.php
```diff
@@ -43,7 +43,7 @@ public function boot()
      */
     protected function configureRateLimiting()
     {
-        RateLimiter::for('api', function (Request $request) {
```
JhumanJ, line 46: Same as the above

client/app.vue
```diff
@@ -48,7 +48,7 @@ export default {
     useOpnSeoMeta({
       title: 'OpnForm',
       description: 'Create beautiful forms for free. Unlimited fields, unlimited submissions. It\'s free and it takes less than 1 minute to create your first form.',
-      ogImage: '/img/social-preview.jpg',
+      ogImage: `${useRuntimeConfig().public.appUrl}/img/social-preview.jpg`,
```
JhumanJ, line 51: If appUrl is properly set with the subfolder path, why do we need this?

client/components/global/Navbar.vue
```diff
@@ -5,7 +5,7 @@
         <div class="flex items-center">
           <NuxtLink :to="{ name: user ? 'home' : 'index' }"
                     class="flex-shrink-0 font-semibold hover:no-underline flex items-center">
-            <NuxtImg src="/img/logo.svg" alt="notion tools logo" class="w-8 h-8"/>
```
JhumanJ, line 8: Same as before - why need if appUrl is properly set with the subfolder path

JhumanJ (changes requested):
Hey thanks a lot for working on this. I left some comments. I don't think we should be adding `${useRuntimeConfig().public.appUrl}` on every file, there must be another way. If we set the appurl properly then we shouldn't need this should we?

JhumanJ:
> Hey thanks a lot for working on this. I left some comments. I don't think we should be adding `${useRuntimeConfig().public.appUrl}` on every file, there must be another way. If we set the appurl properly then we shouldn't need this should we?

@gilesbradshaw tagging on this as I forgot to do it earlier

gilesbradshaw:
Thanks a lot for your comments - I shall work on this when I get the time.  I assume you'd be willing to accept once issues are sorted?

JhumanJ:
> Thanks a lot for your comments - I shall work on this when I get the time. I assume you'd be willing to accept once issues are sorted?

Definitely

PR #364, opened 2024-03-27
fix: opt in to `import.meta.*` properties

danielroe:
This is a very early PR to make this app compatible with [changes we expect to release in Nuxt v5](https://github.com/nuxt/nuxt/issues/25323).

In [Nuxt v3.7.0](https://github.com/nuxt/nuxt/releases/tag/v3.7.0) we added support for `import.meta.*` (see [original PR](https://github.com/nuxt/nuxt/pull/22428)) and we've been gradually updating docs and moving across from the old `process.*` patterned variables.

As I'm sure you're aware, these variables are replaced at build-time and enable tree-shaking in bundled code.
This change affects _runtime_ code (that is, that is processed by the Nuxt bundler, like vite or webpack) rather than code running in Node. So it really doesn't matter what the string is, but it makes more sense in an ESM-world to use `import.meta` rather than `process`.

## Summary by CodeRabbit

- **Refactor**
	- Updated environment detection logic across various components and utilities to use `import.meta` for better handling of server-side and client-side rendering.
	- Improved consistency in environment checks throughout the application, enhancing compatibility and performance.

### DIFF
--- a/client/components/forms/ImageInput.vue
+++ b/client/components/forms/ImageInput.vue
@@ -144,7 +144,7 @@ export default {
   watch: {
     showUploadModal: {
       handler (val) {
-        if (process.server) return
+        if (import.meta.server) return
         document.removeEventListener('paste', this.onUploadPasteEvent)
         if (this.showUploadModal) {
           document.addEventListener('paste', this.onUploadPasteEvent)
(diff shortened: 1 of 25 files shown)


JhumanJ (approved):
Thanks automated Daniel 🙏

danielroe:
Only semi-automated ❤️

PR #371, opened 2024-04-10
feat: disable custom script for  trial users

madassdev:


app/Models/Workspace.php
```diff
@@ -107,6 +108,24 @@ public function getIsProAttribute()
         });
     }
 
+    public function getIsTrialingAttribute()
+    {
+        if (is_null(config('cashier.key'))) {
+            return true;    // If no paid plan so TRUE for ALL
```
JhumanJ, line 114: If no payment, then no one should be trialing

JhumanJ (changes requested):
Small feedback

JhumanJ (approved):

PR #391, opened 2024-04-29
slider input color

madassdev:
## Summary by CodeRabbit

- **Style**
	- Enhanced the visual appearance of the slider input with new styling options, including dynamic color settings for the thumb.


JhumanJ (changes requested):
Hey, some more precisions about this:
- Please make sure that the leftmost part of the cursor is still selected, as it currently is.
- Also it seems like you changed the gray background color, can you please make it be the same as before: a border (same as input border color) and a gray background (same color as disabled input background color)
CleanShot 2024-05-01 at 12 42 32

PR #404, opened 2024-05-07
drag and drop to add  block

madassdev:


JhumanJ (changes requested):
CleanShot 2024-05-07 at 20 15 05@2x

client/components/open/forms/OpenForm.vue
```diff
@@ -451,10 +451,47 @@ export default {
...
+        }
+        // Add the target index on the current page
+        targetIndex += data.added.newIndex;
+        this.workingFormStore.addBlock(data.added.element, targetIndex)
       }
-      if(data.moved){
-        this.workingFormStore.moveField(data.moved)
+      
+      if (data.moved) {
+        const newFields = clonedeep(this.workingFormStore.content.properties)
```
JhumanJ, line 484: Similar as above

client/components/open/forms/OpenForm.vue
```diff
@@ -451,10 +451,47 @@ export default {
     },
     handleDragDropped(data) {
       if(data.added){
-        this.workingFormStore.addBlock(data.added.element, data.added.newIndex)
+        let targetIndex = 0;
+        const newFields = clonedeep(this.workingFormStore.content.properties)
+        const groups = []
+        let currentArray = [];
+        // Group into pages by page break
+        newFields.forEach(item => {
```
JhumanJ, line 459: We already have `fieldGroups` for this on this componenent.. No need to duplicate properties etc
Also, all of this should only run if `currentFieldGroupIndex` > 0

JhumanJ (changes requested):
I think the logic is there, but please revise implementation

JhumanJ (approved):

Issue #394, opened 2024-05-01
Google Sheet - OAuth "client" powered integrations


JhumanJ:
## Context
In https://github.com/JhumanJ/OpnForm/pull/346, we completely refactored integrations. Before you could only have 1 integration of each type (email, slack, webhook etc) for each form, and these integrations were in the form editor. Now there's a dedicated section for integrations, and it's possible to create multiple integration of the same type for each form.

Related guides:
- https://help.opnform.com/en/article/where-are-my-form-integrations-email-submission-confirmation-email-notification-webhook-zapier-slack-discord-ixxg6d/
- https://help.opnform.com/en/article/can-i-trigger-another-service-whenever-my-form-receives-a-submission-webhooks-6nrnfn/

![CleanShot 2024-05-01 at 12 51 49](https://github.com/JhumanJ/OpnForm/assets/11312432/192cda3e-0870-4678-b164-75f52c74d699)

Code-wise, we refactored integrations so that they all share more code, same triggers etc. Making it much easier to add new integrations. We're now missing 2 last pieces to make it super easy to add integrations, both are related to authentication with external services:
- Oauth as a client: to be able to integrate directly with services like google sheet for instance. We need to be able to store the credential of a User for a given application.
- Oauth as a server: to allow automation tools like Zapier for instance to connect to our users account on behalf of them. To create new zaps etc.

The goal of this issue is to handle the first case `Oauth as a client`, and to create our first client integration "Google sheet" while working on this.

## About the GSheet integration
Here's the flow:
- User click on new GSHeet integration
- They select an existing google account they shared with OpnForm, or they can share a new one
- They finalize the creation, which creates a new google sheet in their drive where form submissions will be sent
- On the integration list page, the Gsheet integration cards should have a white button to open the google spreadsheet 
![CleanShot 2024-05-01 at 13 01 03](https://github.com/JhumanJ/OpnForm/assets/11312432/12587a18-2e93-44ec-93b2-992f04a7a2ad)
- Just like for webhooks integrations, errors when trying to add rows to the google sheet should be tracked via the integration events

## Some requirements:
- Users need to be able to connect multiple service accounts of the same service provider (e.g. multiple google accounts)
- For a each integration they should be able to select which account to use (depending on the integration type of course)
- Token/crendentials etc should be [store encrypted](https://laravel.com/docs/11.x/eloquent-mutators#encrypted-casting) in the db, and in general security best practices must be used

## Implementation suggestions
[...]

PR #415, opened 2024-05-21
Google Sheet - OAuth "client" powered integrations

lepikhinb:
Implement https://github.com/JhumanJ/OpnForm/issues/394

Required `.env` variables:

```env
GOOGLE_CLIENT_ID=
GOOGLE_CLIENT_SECRET=
GOOGLE_REDIRECT_URL=http://localhost:3000/settings/connections/callback/google
```


app/Integrations/Google/Google.php
```diff
@@ -0,0 +1,45 @@
+<?php
```
JhumanJ, line 1: Can we try and group these files `app/Integrations` with the rest of the other integration files in `Service/Forms/Integrations`? I'm not a fan of `Service/Forms/Integrations` so we can even move the code from their to here, but I think it would be best to try and keep all of the integrations related code grouped

client/components/open/integrations/components/GoogleSheetsIntegrationActions.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+  <v-button
```
JhumanJ, line 2: Another thought no specified before: since there will be multiple users in workspaces very soon, it would be great here to be able to know what account is connected here, or at least which user created this connection. Maybe that's something we can put here ?
CleanShot 2024-05-22 at 13 52 19@2x

client/components/open/integrations/GoogleSheetsIntegration.vue
```diff
@@ -5,7 +5,25 @@
...
+        v-model="integrationData.oauth_id"
+        name="provider"
+        :options="providers"
+        option-key="id"
+        emit-key="id"
+        :required="true"
+        label="Select Google Account"
+      />
+
+      <v-button
```
JhumanJ, line 19: Here's where I mention that after connecting an account for the first time the "flow" is broken

client/components/open/integrations/GoogleSheetsIntegration.vue
```diff
@@ -5,7 +5,25 @@
...
-      Coming Soon...
+      <select-input
+        v-if="providers.length"
+        v-model="integrationData.oauth_id"
+        name="provider"
+        :options="providers"
+        option-key="id"
+        emit-key="id"
+        :required="true"
+        label="Select Google Account"
```
JhumanJ, line 16: Can you please add a help text "Settings > Connections to add another account"

JhumanJ (changes requested):
Thanks for the great work! I've added small comments to the PR.

Other notes:

- When adding a GSheet integration for the first time, we're prompted to connect our Google account. Once it's done, we're taken back to the connections settings page, instead of going back to the form integration page. Can you please add a mechanism so that the user is at least taken back to the form integrations page?

- Got an exception here:
image

- After creating an integration the gsheet this link doens't work (works after refreshing the page)
CleanShot 2024-05-22 at 11 26 43@2x

- The form structure is dynamic, so users can add/remove columns. With the current integration the index of the column simply is the index of the fields in the form. But as this can be changes (and new columns can be added) this doens't work. We need to find a way to fix this, any suggestion? I'm not sure what's the best way to match form fields (which have unique IDs already) and gSheet columns. Field names can be changed as well so that wouldn't work.. Maybe we could use named cells/ranges and assign field ids as header cell names? 
CleanShot 2024-05-22 at 13 48 15@2x

JhumanJ (approved):

PR #520, opened 2024-08-08
7a137 google auth

madassdev:
## Summary by CodeRabbit

- **New Features**
	- Introduced Google authentication with new "Sign in with Google" options in the login and registration forms.
	- Added a callback component to manage OAuth responses and direct users based on authentication status.
	- Implemented a new environment variable for enhanced OAuth redirect URL configuration.

- **Bug Fixes**
	- Improved error handling during the OAuth process to provide better user feedback.

- **Chores**
	- Cleaned up middleware by removing unnecessary debug logging.

- **Documentation**
	- Updated routing for OAuth connections to enforce POST requests.


app/Http/Controllers/Auth/OAuthController.php
```diff
@@ -44,47 +44,84 @@ public function redirect($provider)
...
 
             return $oauthProvider->user;
         }
 
-        if (User::where('email', $user->getEmail())->exists()) {
-            throw new EmailTakenException();
-        }
 
-        return $this->createUser($provider, $user);
+        if ($provider->getDriver()->canCreateUser()) {
```
JhumanJ, line 94: let's simplify the different IF statements. Instead of creating huge statements, we can negate them. For instance 
```php
if (!$provider->getDriver()->canCreateUser()) { // use negation
  // do something
  return $something;
}

// Rest of the code
```

Also:
- when creating a user: always make sure email is lowercase
- when searching a user by email: use lowercase
- if there's already a user using the same email (but not connected to google account), then respond with an error message explaining that use is already registered with this email and need to use password

client/pages/oauth/callback.vue
```diff
@@ -0,0 +1,92 @@
...
+                    event: 'login',
+                    source: provider
+                })
+            } catch (error) {
+                console.error(error)
+            }
+            router.push({ name: "home" })
+            return
+        } else {
+            console.log('new user')
```
JhumanJ, line 70: Clean this

client/pages/oauth/callback.vue
```diff
@@ -0,0 +1,92 @@
...
+            logEvent("register", { source: provider })
+            router.push({ name: "forms-create" })
+            try {
```
JhumanJ, line 73: Add success message on register

JhumanJ: Success/welcome message

JhumanJ (changes requested):
I haven't tested it but it looks good. Please address the few comments below 🙏

app/Http/Controllers/Auth/OAuthController.php
```diff
@@ -44,47 +44,97 @@ public function redirect($provider)
...
+            return null;
         }
 
-        return $this->createUser($provider, $user);
+        $email = strtolower($socialiteUser->getEmail());
+        $user = User::whereEmail($email)->first();
+        if ($user) {
+            $user->has_registered = true;
+            return $user;
+        } else {
```
JhumanJ, line 111: No need for the `else` statement, you can just put the code inline since there's a return statement

app/Http/Controllers/Auth/OAuthController.php
```diff
@@ -44,47 +44,97 @@ public function redirect($provider)
...
-        $this->guard()->setToken(
-            $token = $this->guard()->login($user)
-        );
-
-        return view('oauth/callback', [
-            'token' => $token,
-            'token_type' => 'bearer',
-            'expires_in' => $this->guard()->getPayload()->get('exp') - time(),
-        ]);
+        try {
```
JhumanJ, line 49: The try catch doesn't need to be around everything - you can just put ut around the first line I believe? 
There should not be exception otherwise?

JhumanJ (changes requested):
More details to polish thanks

app/Http/Controllers/Auth/OAuthController.php
```diff
@@ -44,47 +44,99 @@ public function redirect($provider)
...
+        ]);
+
+        $user->workspaces()->sync([
+            $workspace->id => [
+                'role' => User::ROLE_ADMIN,
+            ],
+        ], false);
+        $user->new_user = true;
+
+        $user->oauthProviders()->create([
```
JhumanJ, line 133: This doesn't work - you need to do the same thing as we do in `OAuthProviderController` using the driver and driver user

[/app/Http/Controllers/OAuth/OAuthProviderController.php#L38-L51](https://github.com/JhumanJ/notionforms/blob/83ddb70051fd87e6c832a2fab4b38d140b70ae55/app/Http/Controllers/OAuth/OAuthProviderController.php#L38-L51)

JhumanJ (changes requested):
It's great - I've already polished the UI a bit.
There's an issue though: OauthProvider model aren't correctly created when creating an account with google. As you can see it's empty (see comments in code)
CleanShot 2024-08-17 at 10 58 35@2x

JhumanJ (approved):

JhumanJ (approved):

PR #568, opened 2024-09-12
Customizable primary color

adriangonzales:
Use `primary` tailwind classes instead of `blue`, allow `primary` to be set with env file


JhumanJ:
Hey @adriangonzales I think this is a good idea - thanks for it. Haven't had time to review yet, any reason for closing it?

PR #687, opened 2025-01-31
Update nginx.conf

niels-website:
Changed undefined variable original_uri to request_uri which should use the original request uri. (nginx container won't start otherwise -- it exits with exit code 1)

## Summary by CodeRabbit

- **Configuration**
	- Updated Nginx configuration to use `$request_uri` instead of `$original_uri` for API routing.


JhumanJ (approved):
Thanks merging!

PR #702, opened 2025-02-13
when duplicate form remove removed_properties

chiragchhatrala:
## Summary by CodeRabbit

- **Bug Fixes**
	- Updated the form duplication process to reset previously removed properties, ensuring duplicated forms start with a clean slate.

### DIFF
--- a/api/app/Http/Controllers/Forms/FormController.php
+++ b/api/app/Http/Controllers/Forms/FormController.php
@@ -186,6 +186,7 @@ class FormController extends Controller
         // Create copy
         $formCopy = $form->replicate();
         $formCopy->title = 'Copy of ' . $formCopy->title;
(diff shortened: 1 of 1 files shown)


api/app/Http/Controllers/Forms/FormController.php
```diff
@@ -186,6 +186,7 @@ public function duplicate($id)
         // Create copy
         $formCopy = $form->replicate();
         $formCopy->title = 'Copy of ' . $formCopy->title;
+        $formCopy->removed_properties = null;
```
JhumanJ, line 189: By default removed properties should be `[]` - I think null may cause some issues

JhumanJ (changes requested):

JhumanJ (approved):

PR #735, opened 2025-03-26
Replace env() usage within app

mgilfillan:
env() will always return the default value when used outside of config files if the config has been cached

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced the application’s handling of its database configuration, ensuring more consistent behavior across environments during form processing and statistics tracking.


JhumanJ (approved):
Thanks!

PR #748, opened 2025-05-01
ESC 542 - Fix email content with bullet list

chiragchhatrala:
## Summary by CodeRabbit

- **New Features**
  - Improved bullet list formatting in the editor, ensuring bullet points display with consistent styling and indentation.

### DIFF
--- a/api/app/Notifications/Forms/FormEmailNotification.php
+++ b/api/app/Notifications/Forms/FormEmailNotification.php
@@ -196,11 +196,7 @@ class FormEmailNotification extends Notification
 
     private function getEmailContent(): string
     {
-        // Convert bullet list to email-safe HTML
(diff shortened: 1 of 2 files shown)


JhumanJ (changes requested):
That's not a good way to solve this issue. Imagine we have style or classes added later this wouldn't work anymore. The solution is rather to fix the issue at the root. Currently RichTextEditor only retuns <ol> instead of sometimes <ul> and sometimes <ol> based on selection.

It seems like this can be fixed by taking the content of the editor the right way (check this comment https://github.com/slab/quill/issues/3957#issuecomment-1945538748) within QuillyEditor.vue#L62-L63. Please try and let me know if this works, but we want to fix this client side.

PR #767, opened 2025-05-27
ESC-575 If operator has no format and no expected_type, it means it doesn't n…

chiragchhatrala:
…eed input

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved condition handling so that fields without an expected value no longer require input, preventing unnecessary validation errors.
- **Tests**
  - Added a test to verify form integration creation with checkbox-based conditional logic.


JhumanJ (approved):

JhumanJ (changes requested):
Please add some tests (and check this comment https://github.com/JhumanJ/OpnForm/pull/767/files#r2108177238)

PR #842, opened 2025-08-07
fix admin page middleware

chiragchhatrala:
## Summary by CodeRabbit

* **Refactor**
  * Updated access control on the admin page to require both authentication and moderator permissions.


JhumanJ:
Not needed as already auth checked in moderator

PR #860, opened 2025-08-25
Admin edit submission

chiragchhatrala:
## Summary by CodeRabbit

- **New Features**
  - Policy-based permission checks for submitting and editing forms.
  - Bulk multi-delete for removing multiple submissions at once.

- **Bug Fixes**
  - Update now returns 404 if a submission is missing.
  - Unauthorized submission attempts are properly blocked by policies.

- **Refactor**
  - Tables, modals, and actions now use mutation-driven flows with centralized loading and alert handling, using in-place paginated cache updates for immediate UI updates.
  - Components no longer emit updated/deleted events to parents (interaction surface simplified).

### DIFF
--- a/api/app/Http/Controllers/Forms/FormSubmissionController.php
+++ b/api/app/Http/Controllers/Forms/FormSubmissionController.php
@@ -104,7 +104,7 @@ class FormSubmissionController extends Controller
     {
         $this->authorize('view', $form);
 
-        $displayColumns = collect($request->columns)->filter(fn($value, $key) => $value === true)->toArray();
+        $displayColumns = collect($request->columns)->filter(fn ($value, $key) => $value === true)->toArray();
 
         // Check if we should process asynchronously
         if ($exportService->shouldExportAsync($form)) {
--- a/api/app/Http/Controllers/Forms/PublicFormController.php
+++ b/api/app/Http/Controllers/Forms/PublicFormController.php
@@ -8,7 +8,6 @@ use App\Models\Forms\Form;
 use App\Http\Resources\FormResource;
 use App\Http\Resources\FormSubmissionResource;
 use App\Jobs\Form\StoreFormSubmissionJob;
-use App\Models\Forms\FormSubmission;
 use App\Service\Forms\Analytics\UserAgentHelper;
 use App\Service\Forms\FormSubmissionProcessor;
 use App\Service\Forms\FormCleaner;
--- a/api/app/Jobs/Form/ExportFormSubmissionsJob.php
+++ b/api/app/Jobs/Form/ExportFormSubmissionsJob.php
@@ -102,23 +102,16 @@ class ExportFormSubmissionsJob implements ShouldQueue
     ): void {
         $exportService = app(FormExportService::class);
         $cacheKey = $exportService->getCacheKey($this->jobId);
-<<<<<<< HEAD
-=======
 
         // Get existing data to preserve created_at
         $existingData = Cache::get($cacheKey, []);
 
->>>>>>> main
         $data = [
             'status' => $status,
             'progress' => $progress,
             'form_id' => $this->form->id,
             'user_id' => $this->userId,
-<<<<<<< HEAD
-            'created_at' => now()->toISOString(),
-=======
             'created_at' => $existingData['created_at'] ?? now()->toISOString(),
->>>>>>> main
             'updated_at' => now()->toISOString(),
         ];
 


chiragchhatrala:
> * Instead of updating the cache value when submission updated, it re-fetches everything on the front-end
> * let's remove the `authorize` from the request, and instead user: a new FormPolicy to check if allowed to answer (used in publicformcontroller), and create FormSubmissionPolicy with a method to check if a user can edit a given policy (if owner) in the form submission controller

Done

client/components/open/forms/components/FormSubmissions.vue
```diff
@@ -81,24 +81,49 @@ const handlePageChange = (page) => {
   setPage(page)
 }
 
-const onUpdateRecord = (_submission) => {
```
JhumanJ: We should manulally have to redo all of this as it's already done in  `useFormSubmissions.js`. The query composable are here to maintain the state and wrap API endpoints. It's great that we have events for record update/delete, but actually we shouldn't to use it here, and rather directly use `useFormSubmissions` in `RecordOperations`. That way state/cache data will be directly update without having to emit the event twice, then update the data etc. That's the whole point of using tanstack query

client/components/open/forms/components/FormSubmissions.vue
```diff
@@ -81,24 +81,49 @@ const handlePageChange = (page) => {
...
+    }
+  )
 }
 
-const onDeleteMultiRecord = (_submissionIds) => {
-  // Invalidate to refetch current page
-  queryClient.invalidateQueries({
-    queryKey: ['forms', props.form.id, 'submissions', 'paginated']
-  })
+const onDeleteRecord = (deletedSubmission) => {
```
JhumanJ: Same for this

api/app/Http/Controllers/Forms/FormSubmissionController.php
```diff
@@ -71,7 +71,8 @@ public function submissions(Form $form)
 
     public function update(AnswerFormRequest $request, Form $form, $submission_id)
     {
-        $this->authorize('update', $form);
+        $submission = FormSubmission::findOrFail($submission_id);
+        $this->authorize('update', $submission);
```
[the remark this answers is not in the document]
JhumanJ: This is a pretty good point, need to check that this submissions belongs to this form.

JhumanJ (changes requested):
Please fix the `RecordOperations`

chiragchhatrala:
> Please fix the `RecordOperations`

Done

client/components/open/components/EditSubmissionModal.vue
```diff
@@ -83,24 +84,32 @@ watch(() => props.show, (newShow) => {
...
+      onError: (error) => {
+        console.error(error)
+        if (error?.data) {
+          alert.formValidationError(error.data)
+        }
       }
-      loading.value = false
-    })
+    }
+  )
```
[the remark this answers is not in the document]
JhumanJ: Please do like the AI suggest, you're currently overriding the default behavior of what happens onsucess, on error. Use mutate async with promise handling

https://github.com/JhumanJ/OpnForm/blob/62d013f5ee5337a4a396bb39a9034e7ebe45413c/client/composables/query/forms/useFormSubmissions.js#L143-L166

api/app/Http/Controllers/Forms/FormSubmissionController.php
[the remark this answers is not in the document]
JhumanJ: @chiragchhatrala you forgot this

client/components/open/components/RecordOperations.vue
```diff
@@ -100,17 +101,23 @@ const onDeleteClick = () => {
...
-        alert.success(data.message)
-      } else {
-        alert.error("Something went wrong!")
+  deleteSubmissionMutation.mutate(
+    { 
+      formId: props.form.id, 
+      submissionId: submission.value.id 
+    },
+    {
+      onSuccess: (data) => {
```
JhumanJ: Same do not use this but instead on vue file when triggering a function use mutate async. If there's a form you can also directly use it to submit like this: 

https://github.com/JhumanJ/OpnForm/blob/62d013f5ee5337a4a396bb39a9034e7ebe45413c/client/composables/lib/vForm/Form.js#L221-L229

client/components/open/tables/OpenTable.vue
```diff
@@ -274,23 +275,26 @@ const onDeleteMultiClick = () => {
...
-        alert.success(data.message)
-      } else {
-        alert.error("Something went wrong!")
+  deleteMultiSubmissionsMutation.mutate(
+    { 
+      formId: props.form.id, 
+      submissionIds: selectedIds.value 
+    },
+    {
+      onSuccess: (data) => {
```
JhumanJ: Same don't use this

JhumanJ (changes requested):
Please check all comments and resolve all conversation, including ai one

chiragchhatrala:
Created another PR #886

PR #969, opened 2025-11-26
Enhance embed code functionality and iframe initialization

chiragchhatrala:
- Updated the embed code in `EmbedCode.vue` to pass an `isFocused` parameter to the `initEmbed` function.
- Modified the `initEmbed` function in `iframe.js` to accept an options object, allowing for conditional behavior based on the `isFocused` state.
- Adjusted iframe initialization logic to prevent unnecessary DOMContentLoaded event handling when in focused mode.

### DIFF
--- a/api/app/Service/AI/Prompts/Form/CheckSpamFormPrompt.php
+++ b/api/app/Service/AI/Prompts/Form/CheckSpamFormPrompt.php
@@ -17,6 +17,12 @@ class CheckSpamFormPrompt extends Prompt
 
         Analyze this form for spam/phishing. Focus on CONTENT RISK, not user characteristics.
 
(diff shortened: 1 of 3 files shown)


client/public/widgets/iframe.js
```diff
@@ -473,12 +473,15 @@
...
  * @type {string}
+ * @param {Object} options
  */
-function initEmbed(formSlug) {
+function initEmbed(formSlug, options = {}) {
   if (!formSlug) {
     return
   }
 
+  const { isFocused = false } = options || {}
```
JhumanJ: instead of isFocused, we should use autoResize since this only controls auto resize - more flexibility for our users

chiragchhatrala: Fixed

PR #977, opened 2025-12-04
ESC-764 - Update OAuthGoogleDriver to include additional Google Sheets scope in…

chiragchhatrala:
… integration permissions


JhumanJ:
We shouldn't expand permissions

PR #983, opened 2025-12-12
ESC-768 - Enhance two-factor authentication handling in login flow

chiragchhatrala:
- Added `handleTwoFactorError` to manage 2FA requirements during login.
- Updated `LoginForm.vue` to incorporate error handling for 2FA scenarios.
- Refactored `useAuthFlow.js` to share 2FA modal state across instances.
- Improved error handling in `useAuth.js` to coordinate 2FA flow effectively.

These changes improve the user experience by providing clearer handling of two-factor authentication during the login process.

### DIFF
--- a/client/components/pages/auth/components/LoginForm.vue
+++ b/client/components/pages/auth/components/LoginForm.vue
@@ -135,7 +135,7 @@ const oAuth = useOAuth()
 const router = useRouter()
 const { login: loginMutationFactory } = useAuth()
 const authFlow = useAuthFlow()
-const { showTwoFactorModal, pendingAuthToken, handleTwoFactorVerified, handleTwoFactorCancel } = authFlow
+const { showTwoFactorModal, pendingAuthToken, handleTwoFactorVerified, handleTwoFactorCancel, handleTwoFactorError } = authFlow
 
 // Feature flags
 const oidcAvailable = computed(() => useFeatureFlag('oidc.available', false))
@@ -243,6 +243,14 @@ const login = () => {
       redirect()
     }
   }).catch((error) => {
+    // Check if this is a 2FA requirement (422 with requires_2fa flag)
+    const twoFactorData = handleTwoFactorError(error)
+    if (twoFactorData) {
+      // This is a 2FA requirement, already handled in onError, just return
+      return
+    }
+    
+    // Handle other errors
     console.log(error)
     if (error.response?._data?.message == "You must change your credentials when in self host mode") {
       redirect()
--- a/client/composables/query/useAuth.js
+++ b/client/composables/query/useAuth.js
@@ -4,7 +4,7 @@ import { initServiceClients } from '~/composables/useAuthFlow'
 
 export function useAuth() {
   const queryClient = useQueryClient()
-  const { handleAuthSuccess, handleManualLogout } = useAuthFlow()
+  const { handleAuthSuccess, handleManualLogout, handleTwoFactorError } = useAuthFlow()
   const { isAuthenticated } = useIsAuthenticated()
 
   // Queries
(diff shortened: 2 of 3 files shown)


JhumanJ:
###  issues

#### 1. Duplicate error handling in `LoginForm.vue`
Problem: `LoginForm.vue` checks for 2FA errors in `.catch()`, but `useAuth.js` already handles them in `onError`. This can cause double handling.

```245:251:client/components/pages/auth/components/LoginForm.vue
  }).catch((error) => {
    // Check if this is a 2FA requirement (422 with requires_2fa flag)
    const twoFactorData = handleTwoFactorError(error)
    if (twoFactorData) {
      // This is a 2FA requirement, already handled in onError, just return
      return
    }
```

Issue: If `onError` calls `handleAuthSuccess`, the mutation may not reject, so `.catch()` may not run. If it does run, this is redundant.

Recommendation: Remove the 2FA check from `LoginForm.vue`'s `.catch()` since `onError` handles it.

#### 2. Error propagation in `useAuth.js`
Problem: `onError` calls `handleAuthSuccess` and returns, but TanStack Query may still treat the mutation as failed.

```73:84:client/composables/query/useAuth.js
      onError: (error, variables) => {
        // Check if this is a 2FA requirement (422 with requires_2fa flag)
        const twoFactorData = handleTwoFactorError(error)
        if (twoFactorData) {
          // Handle auth flow coordination (will show 2FA modal)
          handleAuthSuccess(twoFactorData, variables?.source || 'credentials')
          return
        }
        
        // This is a real error, let it propagate
        throw error
      },
```

Issue: Returning from `onError` doesn't prevent the error from propagating. The mutation will still be in an error state, which can affect loading states and error handling.

Recommendation: Consider handling 2FA in `onSuccess` by checking the response, or document that the mutation may appear failed when 2FA is required.

### Design concerns

#### 3. Shared state outside composable
Good: Moving state outside ensures all instances share the same modal state.

```30:33:client/composables/useAuthFlow.js
// Shared state for 2FA modal - ensures all instances of useAuthFlow share the same state
const showTwoFactorModal = ref(false)
const pendingAuthToken = ref(null)
const pendingAuthContext = ref(null)
```

Consideration: Ensure cleanup on logout/unmount. Verify cleanup in `handleTwoFactorCancel` and `handleManualLogout`.

#### 5. Error structure consistency
The code checks `error.response?._data?.requires_2fa`. Verify this matches your API client's error shape (e.g., Nuxt's `$fetch` vs Axios).


### Recommendations

1. Remove duplicate handling in `LoginForm.vue`:
   ```javascript
   form.mutate(loginMutation).then(() => {
     if (!showTwoFactorModal.value) {
       redirect()
     }
   }).catch((error) => {
     // Remove 2FA check - already handled in onError
     console.log(error)
     if (error.response?._data?.message == "You must change your credentials when in self host mode") {
       redirect()
     }
   })
   ```

2. Consider handling 2FA in `onSuccess`:
   ```javascript
   onSuccess: (tokenData, variables) => {
     // Check if 2FA is required (successful response with 2FA flag)
     if (tokenData.requires_2fa) {
       handleAuthSuccess(tokenData, variables?.source || 'credentials')
       return
     }
     // Normal success flow...
   },
   onError: (error) => {
     // Only handle real errors here
     throw error
   }
   ```
   Note: This depends on whether the backend returns 2FA requirements as 200 or 422.

3. Add cleanup verification: Ensure `handleManualLogout` clears 2FA state:
   ```javascript
   const handleManualLogout = async () => {
     // Clear 2FA state
     showTwoFactorModal.value = false
     pendingAuthToken.value = null
     pendingAuthContext.value = null
     // ... rest of logout
   }
   ```

PR #992, opened 2025-12-23
Add console command to disable two-factor authentication for users

chiragchhatrala:
- Introduced a new command `user:disable-two-factor` to allow administrators to disable two-factor authentication for users via the command line.
- Implemented checks to prevent disabling 2FA for admin users and to ensure that the user exists and has 2FA enabled.
- Added logging functionality to record the action along with the reason provided by the administrator.

These changes enhance the administrative capabilities for managing user security settings directly from the console.

### DIFF
--- a/api/vapor.yml
+++ b/api/vapor.yml
@@ -10,10 +10,10 @@ ignore:
 environments:
   api-production:
     warm: 5
-    memory: 2048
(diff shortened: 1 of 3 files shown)


chiragchhatrala:
> Please add docs :)

Done

PR #1057, opened 2026-03-24
Import form from third-party

chiragchhatrala:


api/app/Service/FormImport/Importers/TypeformImporter.php
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    private function mapFields(array $fields): array
+    {
+        $properties = [];
+
+        foreach ($fields as $field) {
+            $type = $field['type'] ?? 'short_text';
+
+            if (in_array($type, self::COMPOSITE_TYPES)) {
```
JhumanJ, line 256: Importing Typeform into focused mode makes sense, but in that model we should not synthesize `nf-page-break` here. For composite fields this currently leaves empty/trailing pages and over-splits nested groups; I'd keep the flattened fields and let each field become its own focused step instead.

JhumanJ (changes requested):
A few things feel worth tightening before merge:

1. Typeform imports should align with the intended product behavior: import into focused mode, and avoid synthesizing extra page breaks. The current composite-field handling can create empty/trailing pages and over-split group-heavy forms.

2. The import flow would be safer if it stopped short of persisting a new form immediately. Prefilling the editor first, then letting the normal save path run, would give users a preview/confirmation step and reuse the existing validation/cleaning pipeline.

3. This feature needs importer-focused tests before merge. The mapping layer is large and easy to regress. Mocked HTTP fixture tests for each importer, plus a thin controller-level import test, would give much better confidence.

Requesting changes mainly for the Typeform behavior and missing test coverage, and because I think the final save flow should be reconsidered before we lock in the API/UI contract.

api/app/Service/FormImport/Importers/FilloutImporter.php
```diff
@@ -0,0 +1,259 @@
...
+        'LongAnswer' => 'text',
+        'EmailInput' => 'email',
+        'PhoneNumber' => 'phone_number',
+        'NumberInput' => 'number',
+        'URLInput' => 'url',
+        'DatePicker' => 'date',
+        'DateRange' => 'date',
+        'Dropdown' => 'select',
+        'MultiSelect' => 'multi_select',
+        'MultipleChoice' => 'multi_select',
```
JhumanJ: `MultipleChoice` should import as a single-select field, not `multi_select`. `MultiSelect` is already mapped separately, so the current mapping changes the source form's semantics by allowing multiple answers where only one was allowed.

JhumanJ (commented):
Adding one more concrete importer-mapping note we discussed.

chiragchhatrala:
Changes Done

client/pages/forms/create/index.vue
```diff
@@ -98,10 +99,18 @@ onMounted(() => {
   const { hasFeature } = usePlanFeatures()
   form.value = initForm({ workspace_id: workspace.value?.id, no_branding: hasFeature('branding.removal') }, true)
   formInitialHash.value = hash(JSON.stringify(form.value.data()))
```
JhumanJ: `formInitialHash` is captured before imported/template data is merged in below. When an imported form is loaded, the editor will therefore start in a dirty state and can show an unsaved-changes warning before the user edits anything. Recompute the hash after applying imported/template data so the initial prefill state counts as clean.

api/tests/Feature/Forms/FormImportTest.php
```diff
@@ -0,0 +1,544 @@
+<?php
+
+use App\Models\OAuthProvider;
+use Illuminate\Support\Facades\Http;
+
+// ---------------------------------------------------------------------------
+// Controller-level import tests
+// ---------------------------------------------------------------------------
+
+it('returns prefill data for a valid import request', function () {
```
JhumanJ, line 10: The new importer coverage is a good step forward. I think we're still missing assertions for the updated UX contract though: at minimum a test that the prefill payload matches the intended Typeform defaults (for example focused-mode if that's the product decision), plus a regression test for the create-page handoff so imported prefills don't start out dirty immediately.

JhumanJ (commented):
One follow-up after the update: the new prefill flow is much better, but there is still a small create-page regression and I think the test suite should cover it explicitly.

api/app/Service/FormImport/Importers/TypeformImporter.php
```diff
@@ -0,0 +1,242 @@
...
+        );
+    }
+
+    private function fetchFormData(string $formId): array
+    {
+        $response = Http::timeout(15)
+            ->withHeaders([
+                'User-Agent' => 'Mozilla/5.0 (compatible; OpnFormImporter/1.0)',
+            ])
+            ->get(self::API_BASE . $formId);
```
JhumanJ: This request is anonymous, but Typeform's current Create API docs describe `GET /forms/{form_id}` as a `forms:read` endpoint for forms in your account. In practice that means public share URLs do not make this fetch unauthenticated. As written, real imports will hit 401/403 here even though the test suite fakes a 200 response. We need a supported public-data path or a different extraction strategy before advertising Typeform imports.

api/app/Service/FormImport/Importers/GoogleFormsImporter.php
```diff
@@ -0,0 +1,350 @@
...
+
+            return $matches[1];
+        }
+
+        return null;
+    }
+
+    private function fetchForm(string $formId, string $accessToken): array
+    {
+        $response = Http::timeout(15)
```
JhumanJ: We already persist Google `refresh_token`s and refresh stale access tokens for Sheets integrations in `api/app/Integrations/Google/Google.php`, but this importer bypasses that and sends the stored access token straight to the Forms API. After the token ages out, existing Google connections will fail here until the user reconnects manually. Can we refresh on demand (or reuse the existing Google client helper) before surfacing a reconnect error?

JhumanJ (commented):
I took another pass on the updated import flow and found two issues that still look worth fixing before merge. One is on the Typeform backend path itself, and the other is in the Google Forms token lifecycle for already-connected accounts.

JhumanJ:
Before merging, I think this needs broader validation against more complex real-world forms. The current implementation still has too many issues once the source form is not a simple happy path: field order, step/page break reconstruction, label-vs-text detection, standalone checkbox semantics, welcome/success copy, and preserving the original UX for long dropdown flows. I would strongly recommend adding regression coverage for the real examples below (or equivalent fixtures) before merging.

JhumanJ:
Tally example: https://tally.so/r/wMGppm

There are still a few important issues on this form:
- a standalone checkbox is imported as a select-like `multi_select` instead of a real `checkbox`
- question labels such as `Company name` / `Company size` are detected as text blocks, so the following inputs end up as `Untitled`
- the success-page content is not imported cleanly into the resulting form

Could we tighten the label-vs-text heuristic here and add a regression test for this example?

JhumanJ:
Fillout example: https://formsdweddwe.fillout.com/t/88efTSDCqTus

A few things still look off on this one:
- fields are imported in the wrong order
- step/page boundaries are not detected, so the imported form loses the original multi-step structure
- the success-page content is skipped instead of being mapped to `submitted_text`

Nice-to-have: if most steps contain a single interactive field, it may be worth inferring a focused layout automatically, because page-per-question flows become awkward when imported as a classic stacked form.

JhumanJ:
Typeform example: https://3xxbbppenpx.typeform.com/to/DHTu6yrw

This one looks like a very good regression candidate because it exposes three gaps at once:
- the welcome screen text is missing on import
- the thank-you screen content is not recovered
- please make sure real Typeform `dropdown` fields stay dropdowns in OpnForm, especially for long lists like the 16-option country field here; rendering that as a flat option list would be a major UX regression

It would be great to cover welcome screens, thank-you screens, and long dropdowns explicitly in the importer tests.

JhumanJ (changes requested):
Thanks for pushing this forward — the feature is promising, but I don't think it's ready to merge yet.

Requesting changes because the importer still needs to go further on realistic end-to-end coverage and fidelity against more complex source forms. Right now there are still too many gaps on non-trivial examples across providers: field order, page/step reconstruction, label vs text-block detection, standalone checkbox semantics, preserving true dropdown behavior for long option lists, and importing welcome/success content cleanly.

The biggest missing piece is post-submit content: success / thank-you text is still not imported cleanly across providers, and in some cases it appears to be skipped entirely instead of being mapped to `submitted_text`. We also have concrete examples where intro/welcome content is lost, long Typeform dropdowns need to remain dropdowns, Tally checkboxes are imported as select-like fields, and Fillout step boundaries are flattened.

Before merging, I think this should be validated against richer real-world fixtures for each provider and covered with regression tests for the cases called out in the comments above. At the moment the implementation still feels too happy-path-oriented for a feature that needs to preserve the original form structure and UX reliably.

api/app/Service/FormImport/Importers/GoogleFormsImporter.php
chiragchhatrala: Done

api/app/Service/FormImport/Importers/TypeformImporter.php
chiragchhatrala: Done

chiragchhatrala:
Changes Done

PR #1073, opened 2026-04-06
Mention support validation logic

chiragchhatrala:
- Added support for resolving mention references in condition values, allowing single mentions to return raw field values and multiple mentions to resolve to plain-text strings.
- Updated `LogicPropertyValidator` to recognize strings containing 'mention-field-id' as valid types.
- Refactored `FormLogicConditionChecker` to utilize the new mention resolution logic when evaluating conditions.
- Introduced `MentionInput` component for various input types in `ColumnCondition.vue`, ensuring consistent handling of mention fields across the form.
- Enhanced `ConditionEditor.client.vue` and `FormPropertyLogicRule.js` to accommodate mention field logic in property conditions.

These changes improve the handling of mentions in form logic, enhancing user experience and data integrity.

Fixed https://github.com/OpnForm/OpnForm/discussions/1046


JhumanJ:
This doesn’t feel complete yet.

The feature changes condition evaluation in both the client and server, but the PR doesn’t add any tests for the new mention-based behavior. For a change this invasive, I’d expect regression coverage at least around:

- single bare mentions resolving to raw scalar values
- mixed-content / multi-mention values resolving to plain text
- computed-variable mentions working through the real backend call paths
- non-scalar mentions (`multi_select`, `files`, `matrix`, etc.) not crashing string operators

I also found two blocking issues in the current implementation:

1. A single bare mention can now resolve to an array/object, but the backend text comparators still pass `condition['value']` directly into `Str::contains`, `str_starts_with`, and `str_ends_with`. That will raise a `TypeError` for non-string values instead of returning `false`.

2. The PR adds computed-variable support through `FormLogicConditionChecker::conditionsMetWithForm()`, but the real backend callers still use `conditionsMet()`. So custom validations, integration logic, and property visibility/required evaluation won’t actually see computed-variable mentions at runtime.

Because of those gaps, I don’t think this is ready to merge yet. I’d want the backend call sites fixed and the corresponding tests added before approving.

chiragchhatrala:
Changes Done

JhumanJ (approved):

JhumanJ (approved):

PR #1075, opened 2026-04-06
Copilot/add success page custom html

Wiido13:


JhumanJ:
Thanks a lot for taking the time to work on this and open the PR.

We do want to support more flexible success-page customization, and field-to-field comparisons are also something we’ve been thinking about. That said, we can’t merge this as-is right now.

The main issue is that this implementation introduces raw custom HTML on the public success page without the same sanitization and guardrails we use elsewhere, which creates security and product-boundary concerns. It also bundles in unrelated form-logic changes, and there’s at least one user-facing regression in the current diff (the post-submit PDF download action disappears).

So this is more than a small feature addition for us — it needs to be designed and shipped with clearer safety constraints, gating, and coverage.

Really appreciate the contribution and the time that went into it. We’re going to close this for now, but the underlying idea is still on our radar and we may come back to it with a narrower and safer implementation later.

PR #1080, opened 2026-04-14
refactor: update Sentry user context handling

chiragchhatrala:
Replaced the use of `useAuth()` and `vue-query` in the Sentry `beforeSend` function with a direct reference to the `authStore`. This change simplifies the user context retrieval by checking the token and user directly from the store, enhancing performance and reliability in the Sentry integration.

Sentry-6750033165

### DIFF
--- a/client/composables/useAuthFlow.js
+++ b/client/composables/useAuthFlow.js
@@ -16,15 +16,10 @@ export const useIsAuthenticated = () => {
 }
 
 /**
- * Mirror user into the auth store and initialize vendor clients (Amplitude, Crisp).
+ * Initialize vendor clients (Amplitude, Crisp) from the current user.
  * Safe to call from middleware or anywhere outside setup context.
  */
 export const initServiceClients = (userData) => {
-  const authStore = useAuthStore()
-  if (userData !== undefined) {
-    authStore.updateUser(userData ?? null)
-  }
-
   if (import.meta.server) return
   if (!userData) return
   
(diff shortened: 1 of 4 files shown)


JhumanJ (changes requested):
I double-checked this against the current auth flow in the repo before leaving this review.

The new `beforeSend` logic depends on `authStore.user`, but that field is not hydrated anywhere in the normal client auth lifecycle right now:

- `useAuthStore()` initializes `user` to `null`.
- `initStore()` only restores `token` / `admin_token` from cookies.
- login / register / oauth success write the current user into the Vue Query `['user']` cache.
- `01.check-auth.global.js` reads that cache and calls `initServiceClients(userData)`, but does not copy the user into the auth store.

So after this change, authenticated sessions will typically have `authStore.token` set while `authStore.user` remains `null`, which means `Sentry.setUser()` will stop getting `id` / `email` through this path.

Please either keep reading from the existing user source in a context-safe way, or hydrate `authStore.user` as part of login/bootstrap before switching Sentry to the store.

chiragchhatrala:
Changes Done

JhumanJ:
Thanks for updating this — the original `authStore.user` hydration concern looks addressed, but I found a new blocker in the follow-up change.

`initServiceClients()` now calls `authStore.updateUser(userData ?? null)`, while `authStore.updateUser()` still calls `initServiceClients(payload)`. That creates an immediate recursion loop:

`initServiceClients(userData)` -> `authStore.updateUser(userData)` -> `initServiceClients(userData)`

So any auth/bootstrap path that reaches `initServiceClients()` will keep recursing until the stack blows up. The middleware still calls `initServiceClients(userData)` after fetching the current user, so this should be hit on normal authenticated bootstrap as well.

I think the clean fix is to keep the responsibilities one-way:
- `authStore.updateUser()` should only update store state
- `initServiceClients()` should only sync vendor clients
- call both explicitly from login/bootstrap where user data is available

JhumanJ (approved):

PR #1091, opened 2026-04-27
[codex] Fix submission columns, export, and checkbox labels

JhumanJ:
## Summary

Fixes two user-reported form issues:

- Submissions table could show only selection/action columns and export could start with an empty column payload.
- Clicking the visible label next to a checkbox did not toggle the checkbox.

## Root cause

Dashboard form detail queries shared cache keys with public form detail queries. If a public/lightweight form response populated the cache before the private dashboard response, the submissions page could render without the form `properties`, leaving the column manager empty while submissions rows still loaded.

For checkbox fields, the visible label rendered next to the checkbox was not associated with the checkbox input. The only native label linked to the checkbox lived inside `VCheckbox`, but it had no visible text in this layout.

## Changes

- Scope form detail query keys by public/private access and prevent private detail queries from falling back to public endpoints while auth is initializing.
- Normalize table column sources so `properties` and `removed_properties` work as arrays or object maps.
- Add an export fallback that sends all form property ids plus `created_at` when column visibility is empty.
- Associate the visible `CheckboxInput` label with the checkbox input via `native-for`.
- Add a component regression test for checkbox label/input association.

## Validation

- `npm run lint` in `client/`
- `NODE_OPTIONS='--no-experimental-webstorage' npm run test -- --run test/nuxt/checkbox-input.spec.ts test/unit/usePlanFeatures.test.ts test/unit/FormSubmissionFormatter.test.ts` in `client/`
- `./vendor/bin/pest tests/Feature/Forms/FormSubmissionExportTest.php tests/Feature/Submissions/SubmissionTest.php` in `api/`

## Notes

Local DB does not contain the reported customer form slug `liste-dattente-peps-ftxidj`, so exact visual verification on that form was not possible locally.

### DIFF
--- a/client/components/forms/core/CheckboxInput.vue
+++ b/client/components/forms/core/CheckboxInput.vue
@@ -32,6 +32,7 @@
             :label="label"
             :required="required"
             :theme="theme"
+            :native-for="id ? id : name"
           />
         </slot>
         <slot
--- a/client/components/open/forms/FormExportModal.vue
+++ b/client/components/open/forms/FormExportModal.vue
@@ -132,6 +132,25 @@ const animationFrame = ref(null)
 const lastApiCall = ref(null)
 const progressStartTime = ref(null)
 
+const normalizeColumnList = (columns) => {
+  if (Array.isArray(columns)) return columns
+  if (columns && typeof columns === 'object') return Object.values(columns)
+  return []
+}
+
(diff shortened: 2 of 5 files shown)


JhumanJ:
Closing in favor of a clean PR based on main.

PR #1105, opened 2026-05-14
ESC-873 = feat: implement handling of special characters in select options and …

chiragchhatrala:
…add corresponding tests

- Removed the previous escaping logic for select and multi-select options.
- Added a new test suite to validate form submissions with special characters, including tabs, backslashes, newlines, unicode, commas, and quotes.
- Ensured that form submissions correctly handle and store these special characters.

### DIFF
--- a/api/tests/Feature/Forms/SelectOptionSpecialCharsTest.php
+++ b/api/tests/Feature/Forms/SelectOptionSpecialCharsTest.php
@@ -1,6 +1,5 @@
 <?php
 
-use App\Models\Forms\Form;
 
 describe('select options with special characters', function () {
     beforeEach(function () {
--- a/api/tests/Feature/Forms/SelectOptionSpecialCharsTest.php
+++ b/api/tests/Feature/Forms/SelectOptionSpecialCharsTest.php
@@ -234,9 +234,9 @@ describe('select options with special characters', function () {
         expect($submission->data['multi_special'])->toBe(["Tab\there", 'Back\\slash', 'Comma, inside']);
     });
 
-    it('can submit select option by id when option contains special characters', function () {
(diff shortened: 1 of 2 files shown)


api/tests/Feature/Forms/SelectOptionSpecialCharsTest.php
```diff
@@ -0,0 +1,356 @@
...
+                    ['id' => 'opt2', 'name' => 'Normal'],
+                ],
+            ],
+        ];
+
+        $this->form->properties = array_merge($this->form->properties, [$selectField]);
+        $this->form->save();
+
+        $formData = $this->generateFormSubmissionData($this->form, [
+            'select_by_id' => "Special\tOption — Опция",
```
JhumanJ: This test says it submits the option by id, but the payload is the option name again. Please submit `opt1` here and assert the stored value, or rename the test so it does not claim id-based coverage.

JhumanJ (changes requested):
Thanks for the focused fix here. I’m requesting changes because there is still a backslash edge case that this PR does not cover yet.

1. `AnswerFormRequest` still validates select values via `Rule::in($this->getSelectPropertyOptions($property))`. Laravel stringifies that rule and parses it with CSV/backslash escaping, so an option name that ends with a backslash is still rejected. I verified this directly in the app validator with an allowed value of `"Ends\\"`, which fails validation. Please cover and fix trailing-backslash option names, not only internal backslashes like `C:\\Users\\test`.

2. The “submit select option by id” test currently submits the option name again rather than `opt1`, so it does not exercise id-based submission. Please either rename the test or change the payload to submit the id and assert the stored value.

PR #1113, opened 2026-05-20
feat(api): add TRUSTED_PROXIES env configuration support

lilsnibbi:
This update introduces the `TRUSTED_PROXIES` environment variable to allow dynamic configuration of trusted proxies. The variable is mapped through `config/app.php` to ensure compatibility with Laravel's config caching in production. The middleware also now supports comma separated IP strings for multiple proxies or a wildcard `*` (default to ease user config).

This change adds support for deployments behind proxies, allowing the "collect IPs" option to function correctly.


JhumanJ:
Thanks for the contribution! I don’t think this is safe to merge. On our Laravel version, `TrustProxies` does not have a parent constructor, so `parent::__construct($config)` will fatal at runtime. Also, defaulting `TRUSTED_PROXIES` to `*` in Docker is too broad for a safe default because it makes forwarded IP headers trusted by default.

PR #1119, opened 2026-05-22
Update BlockRenderer.vue to improve mobile device UX

doubledods:
Adding **tel**, **email**, and **url** types to input fields for better mobile phone user experience.

- This change improves the mobile phone experience for end users. Their phone keyboard will pop up with the appropriate settings for the tel, email, and url inputs. For example, on email fields, mobile phone keyboards will show the @ symbol on the first page of the keyboard.


JhumanJ:
Thanks for this, it's a great change, but it should be using inputmode instead of input type which can break validation and change rules which mayb break existing forms. CLosing for now please feel free to reopen.

JhumanJ (approved):
Looks good. This is a small UX improvement for mobile keyboards via native input types. Product checks pass; the remaining failed check appears to be the PR notification webhook rather than the app. Please rebase and clean up indentation if convenient.

PR #1121, opened 2026-05-24
fix: ensure uploaded asset URLs use HTTPS scheme

Zekbot001:
## Fixes
Closes #1068

## Changes
When uploading form assets (favicon, images), \
oute()\ generates a URL using \APP_URL\ config. In self-hosted deployments where \APP_URL\ uses \http://\, the returned URL causes browsers to block the favicon as mixed content.

**Before:** \http://example.com/forms/assets/...\
**After:** \https://example.com/forms/assets/...\

## Testing
- Verified the change compiles with PHP syntax checks
- The fix is a 1-line change: \
oute()\ output is now forced to use \https://\ scheme
- No backend logic changes, only URL generation


JhumanJ:
Thanks - closing because of lack of tests, bur more importantly i don't believe this is the ideal solution

PR #1124, opened 2026-05-27
Update UserInviteController authorization and enhance UserInvite mode…

chiragchhatrala:
…l security

- Changed authorization method in UserInviteController from 'view' to 'adminAction' to restrict invite listing to admin users.
- Added 'token' to the hidden attributes in UserInvite model to prevent exposure in API responses.
- Introduced new feature tests to ensure proper authorization and token redaction in invite listings.


JhumanJ:
Thanks for tightening this up. One follow-up before merge: please update the API docs and tests to match the new invite-listing contract. `GET /open/workspaces/{workspace}/invites` now goes through `adminAction`, so it is no longer just a `workspace-users-read`/member endpoint, and the expected admin/token-ability behavior should be documented and covered.

Also, the Members settings UI is visible to non-admin workspace members and currently always fetches invites. With this PR those users will get a 403 just by opening the Members tab, even if listing accepted members is still allowed. Please gate the invite query/actions behind `workspace.is_admin` or handle that state intentionally.

JhumanJ (changes requested):
Requesting changes for two follow-ups before merge:

1. Please update the API docs and test coverage to match the new invite-listing authorization contract. The endpoint now uses `adminAction`, so it no longer behaves like the documented `workspace-users-read`/member endpoint.

2. Please handle the frontend Members tab regression for non-admin workspace members. The tab still loads for them and always fetches invites, so this PR makes those users hit a 403 just by opening member settings unless invite fetching/actions are gated behind admin access or handled intentionally.

chiragchhatrala:
Fixed

JhumanJ (approved):
Looks good. The invite-listing contract now requires admin access, invite tokens are hidden, the Members UI gates invite fetching/actions for admins, and docs/tests were updated. Please rebase/rerun CI to clear the stale state before merge.

PR #1128, opened 2026-05-27
Enhance legacy Hashid handling in PublicFormController

chiragchhatrala:
- Refactor submission ID decoding logic to check for existing submissions with UUIDs and return a 404 error if found.
- Add tests to ensure proper handling of Hashid updates based on submission UUID presence, allowing updates for legacy submissions without UUIDs while rejecting those with UUIDs.

### DIFF
--- a/api/app/Http/Controllers/Forms/PublicFormController.php
+++ b/api/app/Http/Controllers/Forms/PublicFormController.php
@@ -5,6 +5,7 @@ namespace App\Http\Controllers\Forms;
 use App\Http\Controllers\Controller;
 use App\Http\Requests\AnswerFormRequest;
 use App\Models\Forms\Form;
+use App\Models\Forms\FormSubmission;
 use App\Http\Resources\FormResource;
 use App\Http\Resources\FormSubmissionResource;
 use App\Jobs\Form\StoreFormSubmissionJob;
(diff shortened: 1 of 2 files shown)


JhumanJ (changes requested):
[P1] Block legacy `submission_id` updates when editable submissions are disabled

I think this still leaves a public edit-control bypass in the legacy Hashid path.

This PR correctly rejects a Hashid when it resolves to a submission that already has `public_id`, but for true legacy submissions (`public_id = null`) we still do:

```php
$submissionData['submission_id'] = $numericId;
```

That happens before any check that the form/workspace currently allows editable submissions. From there, `StoreFormSubmissionJob::extractMetadata()` treats the numeric `submission_id` as a client-provided update target, and `storeSubmission()` updates that existing row via `findOrFail($this->submissionId)`.

The read side already enforces a different contract: `fetchSubmission()` returns `403` when `editable_submissions` is disabled. So a form can block fetching/editing a submission through the normal edit flow, while still accepting a POST with a legacy Hashid and updating the underlying submission. I confirmed this locally with a failing regression probe: create a public form with `editable_submissions = false`, create a legacy submission with `public_id = null`, POST to `forms.answer` with `submission_id = Hashids::encode($submission->id)`, and the existing row is updated instead of being rejected or treated as a new submission.

I don't think this is intended behavior. If a request supplies `submission_id` or `submission_hash`, identifier resolution should only happen when the workspace has editable submissions access and the form has `editable_submissions = true`. Otherwise the endpoint should reject the request, probably with 403/404, before passing a numeric `submission_id` into `StoreFormSubmissionJob`.

Suggested coverage: add a regression test for `editable_submissions = false` + legacy `public_id = null` + Hashid in `submission_id`, asserting that the original submission is not updated.

chiragchhatrala:
Fixed

JhumanJ (approved):
Looks good now. The updated identifier handling prevents legacy hashid/UUID requests from updating existing submissions when editable submissions are disabled, while preserving the intended legacy and partial-submission paths. I re-ran the targeted SubmissionIdentifierTest suite locally and CI is green.

PR #1129, opened 2026-05-27
Enhance AnswerFormRequest validation for partial submissions

chiragchhatrala:
- Update validation logic to only skip checks for legitimate partial submissions, ensuring that both the form and workspace settings allow for partial submissions.
- Add tests to validate required fields when 'is_partial' is sent without proper configuration, ensuring robust error handling for unsupported scenarios.


api/app/Http/Requests/AnswerFormRequest.php
```diff
@@ -46,8 +46,12 @@ private function getFieldMaxFileSize($fieldProps)
      */
     public function rules()
     {
-        // Skip validation if this is a partial submission
-        if ($this->has('is_partial')) {
+        // Only skip validation for legitimate partial submissions (form + plan must allow it)
+        if (
+            $this->has('is_partial')
```
JhumanJ: `has('is_partial')` is still true for `is_partial: false` or `0`. On forms that support partial submissions, that skips all validation even though the controller later treats the request as a completed submission. Please check a truthy boolean value instead and add a regression test for `is_partial => false`.

JhumanJ (changes requested):
Thanks for tightening the partial-submission validation path. I am requesting changes because the current condition is still bypassable for completed submissions.

`has('is_partial')` checks presence rather than truthiness. On a Business form with partial submissions enabled, a request like `{"is_partial": false}` still returns empty validation rules, while the controller later treats it as a completed submission. That allows required fields to be skipped. Please switch this to a truthy boolean check and add a regression test for `is_partial => false` on a required-field form.

chiragchhatrala:
Done

JhumanJ (changes requested):
Thanks for the update. The original bypass with JSON `{ "is_partial": false }` looks fixed, and the added regression test covers that case.

There is still a boolean normalization mismatch between validation and storage. `AnswerFormRequest` now uses Laravel's boolean semantics, but `PublicFormController` later reads `$request->get('is_partial')` directly. A public request with `is_partial: "false"` is false for validation, but truthy in PHP when the controller stores the submission, so it can be saved as a partial draft.

Please normalize once with `$request->boolean('is_partial')` in the controller and pass that value through, and add a regression test for the string `"false"` case.

chiragchhatrala:
Done

JhumanJ (approved):

PR #1131, opened 2026-05-27
fix: display existing signature image when viewing submissions

Zekbot001:
## Summary
- Fixes #1063 - Signature images now display when viewing existing submissions instead of showing an empty canvas
- The `SignatureInput` component now detects the API's array format `[{file_url, file_name}]` and converts it for the `UploadedFile` component

## Changes
- Added a `compVal` watcher with `immediate: true` in `SignatureInput.vue`
- When the field is disabled (viewing mode) and contains existing signature data in the API array format, the watcher constructs a compatible `file` object so the signature image renders

## Test plan
- [x] Code change is minimal (15 lines added) and scoped to the display-only path (guarded by `this.disabled`)
- [x] Does not affect signature input/capture flow for new form submissions
- [ ] Manual verification: view a submission with a signature field and confirm the signature image appears in the form view
- [ ] Manual verification: create a new form submission with a signature and confirm drawing/upload still works

## Limitations
- The email notification "Invalid signature" link issue mentioned in #1063 requires separate backend investigation (likely related to signed URL handling in email clients)
- Only the first signature in the array is displayed (matching the single-signature design of the component)

Closes #1063

## Bounty
This PR addresses a bug eligible for OpnForm's bounty program (via Algora).

### DIFF
--- a/client/components/forms/heavy/SignatureInput.vue
+++ b/client/components/forms/heavy/SignatureInput.vue
@@ -107,6 +107,21 @@ export default {
       handler(file) {
         this.compVal = file?.url || null
       }
+    },
+    compVal: {
+      immediate: true,
+      handler(val) {
+        // When viewing an existing submission, the API returns signature data
+        // as [{file_url: "...", file_name: "..."}]. Convert to the format
+        // expected by UploadedFile so the signature image is displayed.
+        if (this.disabled && val && Array.isArray(val) && val.length > 0 && val[0].file_url) {
+          this.file = {
+            file: { name: val[0].file_name || 'signature.png' },
+            url: val[0].file_url,
+            src: val[0].file_url
+          }
+        }
+      }
     }
   },
 


client/components/forms/heavy/SignatureInput.vue
```diff
@@ -107,6 +107,21 @@ export default {
       handler(file) {
         this.compVal = file?.url || null
       }
+    },
+    compVal: {
```
JhumanJ, line 111: This sets `file` when the incoming value is the API signature array, but it never clears `file` when `compVal` changes to an empty or non-compatible value. If the component is reused across submissions, a previous signature can stay visible.

JhumanJ (changes requested):
The goal is useful, and the narrow change is close. I am requesting changes for one stale-state edge case before this ships.

The new watcher sets `file` when the incoming value matches the API signature array shape, but it never clears `file` when the value changes to empty or to a non-compatible value. If the component instance is reused while navigating between submissions, an old signature can remain visible. Please add a reset branch and cover it with a small test or manual verification note.

PR #1134, opened 2026-05-28
Add SDK bridge messaging functionality for iframe communication

chiragchhatrala:
- Introduced a new `sdkBridgeMessaging.js` file to handle messaging between the parent window and iframe, including functions for resolving trusted origins and reading SDK tokens from URLs.
- Updated `useSdkBridge.js` to utilize the new messaging functions, enhancing the communication flow and security checks for commands and responses.
- Enhanced the `opnform-sdk.js` to support SDK token generation and handshake acknowledgments, improving the integration of embedded forms.
- Added unit tests to ensure the reliability of the new messaging features and their interactions with iframe components.

These changes significantly improve the SDK's capabilities for secure and efficient communication between the parent and iframe contexts.


client/lib/sdk/sdkBridgeMessaging.js
```diff
@@ -0,0 +1,152 @@
...
+export function resolveInitialTrustedOrigin(isIframe) {
+  if (!isIframe || typeof window === 'undefined') return null
+
+  try {
+    const params = new URLSearchParams(window.location.search)
+    const parentOrigin = params.get('_sdkParentOrigin')
+    if (parentOrigin) {
+      return new URL(decodeURIComponent(parentOrigin), window.location.href).origin
+    }
+  } catch (_) { /* ignore */ }
```
JhumanJ: The `_` parameter is unused and currently fails ESLint. Please switch these blocks to `catch { ... }` or otherwise satisfy the lint rule.

client/public/widgets/opnform-sdk.js
```diff
@@ -127,34 +144,113 @@
...
+      return new Promise((resolve) => {
+        this._handshakeResolve = resolve
+        this._sendHandshake()
+
+        this._handshakeRetryTimer = setInterval(() => {
+          if (this._handshakeComplete) return
+          this._sendHandshake()
+        }, 250)
+
+        setTimeout(() => {
```
JhumanJ, line 178: On timeout this marks the handshake as complete and resolves, but the iframe may never have registered `_sdkToken`. Commands then proceed with a token the child rejects. Please treat timeout as handshake failure or keep retrying with a clear error instead of silently succeeding.

JhumanJ (changes requested):
I am requesting changes here because there are two concrete blockers:

1. The client CI is currently red because the new catch parameters are unused. Please switch those blocks to the project-accepted lint style, for example `catch { ... }`.
2. The SDK handshake timeout is treated as success. For existing or slow iframes without a registered `_sdkToken`, `_ensureHandshake()` resolves after the timeout and commands are then sent with a token the child never accepted. That turns the handshake failure into later invalid-token/command-timeout behavior. Please fail or keep retrying the handshake explicitly instead of marking it complete on timeout.

chiragchhatrala:
Done

JhumanJ (changes requested):
Thanks for the update. The two previous blockers are fixed: the lint issue is gone, and command calls now reject when the SDK handshake times out instead of treating that timeout as success.

I am still requesting one small follow-up before merge: `_startHandshake()` now creates a promise that can reject during passive iframe discovery, but that promise is not always consumed. Existing, slow, or legacy iframes that do not acknowledge within the timeout can trigger a global unhandled rejection just from loading the SDK.

Please either attach an explicit no-op catch to the passive startup handshake, or split passive discovery from the command-gating handshake so only SDK method calls reject to callers.

JhumanJ (changes requested):
Thanks for the update. The two previous blockers are fixed: the lint issue is gone, and command calls now reject when the SDK handshake times out instead of treating that timeout as success.

I am still requesting one small follow-up before merge: `_startHandshake()` now creates a promise that can reject during passive iframe discovery, but that promise is not always consumed. Existing, slow, or legacy iframes that do not acknowledge within the timeout can trigger a global unhandled rejection just from loading the SDK.

Please either attach an explicit no-op catch to the passive startup handshake, or split passive discovery from the command-gating handshake so only SDK method calls reject to callers.

JhumanJ (changes requested):
Thanks for the update. The two previous blockers are fixed: the lint issue is gone, and command calls now reject when the SDK handshake times out instead of treating that timeout as success.

I am still requesting one small follow-up before merge: `_startHandshake()` now creates a promise that can reject during passive iframe discovery, but that promise is not always consumed. Existing, slow, or legacy iframes that do not acknowledge within the timeout can trigger a global unhandled rejection just from loading the SDK.

Please either attach an explicit no-op catch to the passive startup handshake, or split passive discovery from the command-gating handshake so only SDK method calls reject to callers.

chiragchhatrala:
Done

JhumanJ (approved):

PR #1135, opened 2026-05-28
Feat: custom form field IDs

bigheadsmith:
Based off of feedback comment #1049 

- Added input field to form field settings that shows the form ID that is currently automatically created
- Allows editing the field ID
- Checks for uniqueness and gives validation errors on public if conditions are not met

**DISCLOSURE**: Claude was used to create this 

image
image
image

### DIFF
--- a/api/app/Rules/FormPropertiesRule.php
+++ b/api/app/Rules/FormPropertiesRule.php
@@ -96,6 +96,22 @@ class FormPropertiesRule implements ValidationRule, ValidatorAwareRule
             }
         }
 
+        // Enforce unique IDs across all properties
+        $ids = array_filter(array_column($value, 'id'), fn($id) => $id !== null && $id !== '');
+        $duplicateIds = array_filter(array_count_values($ids), fn($count) => $count > 1);
+
+        if (!empty($duplicateIds)) {
+            foreach ($value as $index => $property) {
+                if (isset($property['id']) && $property['id'] !== '' && isset($duplicateIds[$property['id']])) {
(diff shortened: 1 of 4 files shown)


filipesmedeiros:
It seems to me that there is no default ID when the field is created? We should probably have a default ID, with the same generation logic that is currently used, and use that default ID as a default value for the input? This is a very useful PR! :)

bigheadsmith:
> It seems to me that there is no default ID when the field is created? We should probably have a default ID, with the same generation logic that is currently used, and use that default ID as a default value for the input? This is a very useful PR! :)

Unsure what you mean exactly.
When I add a new field they are created with a unique ID like below
Recording 2026-06-01 104528

filipesmedeiros:
Ahhh sorry @bigheadsmith. That is exactly what I meant :) Thank you

JhumanJ:
HI there, thanks for the PR. I think allowing edits of fields ids can introduce a lot of issues/side-effects: people changing ids and breaking automations, or even IDs clash. Also displaying the ID like this seems to be pushing down the other options quite a lot. I'll close this for now but I'd be more than happy to discuss about how we can make field ids more accessible and user/friendly

filipesmedeiros:
@JhumanJ what about a read-only ID, but that you can optionally set on create. For example, a global setting called "Custom IDs" that enables custom IDs like this PR does, but becomes read-only on first save. If the global setting (a boolean) is `false` then, the behaviour is exactly the one as it is currently.

PR #1136, opened 2026-05-29
Enhance PDF image handling and form submission features

chiragchhatrala:
- Updated PdfImageResolver to support fetching images from both local storage and remote URLs, improving flexibility in image sourcing.
- Introduced PdfSafeImageFetcher for safe remote image fetching with validation against SSRF attacks.
- Enhanced form submission settings to integrate PDF template capabilities, allowing for improved PDF generation and customization.
- Added comprehensive tests to validate new image handling and PDF integration features, ensuring robustness and reliability.

These changes collectively enhance the user experience and functionality related to PDF generation and image management in forms.


JhumanJ (changes requested):
Changes requested.

Three issues need fixing before approval:

1. P1 security: `PdfSafeImageFetcher.php` reuses `PublicWebhookUrl::assertSafe()` / `requestOptions()`. If `opnform.webhooks.allow_private_urls` is enabled, that class allows private IPs. Since a submitted image or URL value can be rendered into a PDF, this can become SSRF against private network targets on instances that enabled private webhook URLs. Please use a dedicated PDF image fetch policy that remains public-only by default, or prevent submitter-controlled values from triggering private fetches.

2. P2 robustness: `MAX_BYTES` is checked only after `Http::get()` and after reading the full response body. A public endpoint can return a huge response and consume memory or bandwidth before rejection. Please pre-check `Content-Length` and/or stream with a hard download limit.

3. P2 functional: `FilenameUrlEncoder::isPlausibleFilename()` excludes `!` and `*`, while `S3KeyCleaner` and existing behavior support those characters. Encoded filenames containing these characters can fail `isEncoded()` and then not be decoded/resolved for PDFs. Please align the whitelist with the actual allowed filename characters and add coverage for `!` / `*`.

chiragchhatrala:
Done

api/app/Service/Pdf/PdfSafeImageFetcher.php
```diff
@@ -0,0 +1,107 @@
...
+
+            if (!$response->successful()) {
+                return null;
+            }
+
+            if (!$this->responseHeadersAreSafe($response)) {
+                return null;
+            }
+
+            $body = $response->body();
```
JhumanJ: This limit is checked after `Http::get()` and after reading the full body. A chunked response without `Content-Length` can still consume memory/bandwidth before rejection. Please stream with a hard byte cap or use a Guzzle sink/limit that aborts during download.

api/app/Service/Pdf/PdfImageResolver.php
```diff
@@ -42,37 +56,119 @@ public function resolveContent(string $imageValue): ?string
...
+        if ($this->isUrl($imageValue) && !$this->isLocalAssetUrl($imageValue)) {
+            return [];
+        }
+
         $candidates = [];
-        $normalized = trim($imageValue);
 
-        if ($normalized === '') {
-            return $candidates;
+        foreach ($this->extractFileNames($imageValue) as $fileName) {
```
JhumanJ, line 83: For a local `/forms/assets/...` URL, this tries submission storage before static form assets when a form is present. A basename collision can render a submission upload instead of the intended asset. Local asset URLs should prefer or restrict to `ASSETS_UPLOAD_PATH`.

api/app/Service/Pdf/PdfImageResolver.php
```diff
@@ -42,37 +56,119 @@ public function resolveContent(string $imageValue): ?string
...
+    {
+        return $this->remoteFetcher ?? new PdfSafeImageFetcher();
+    }
+
+    private function isUrl(string $value): bool
+    {
+        return filter_var($value, FILTER_VALIDATE_URL) !== false;
+    }
+
+    private function isLocalAssetUrl(string $url): bool
```
JhumanJ: This classifies any URL whose path ends in `/forms/assets/{file}` as local, regardless of host. A public CDN URL with the same path will be skipped as remote and then return null or collide with local storage. Please also validate the host/scheme against the app asset origin.

JhumanJ (changes requested):
The dedicated public-image policy is a good direction, and it resolves the earlier private-URL configuration issue. I am still requesting changes because a few production edge cases remain:

1. The max-size guard is checked after the full response body is downloaded, so a chunked response without `Content-Length` can still consume memory/bandwidth before rejection. Please stream with a hard byte cap or use a sink/limit that aborts during download.
2. Local `/forms/assets/...` URLs can resolve through submission storage before static form assets when a form is present, so a basename collision may render the wrong image. Asset URLs should prefer or restrict to the asset upload path.
3. The local-asset URL check only inspects the path. A remote URL with the same `/forms/assets/...` path can be treated as local; please also validate the expected host/scheme or separate reserved-path handling from origin checks.

chiragchhatrala:
Done

JhumanJ (changes requested):
Thanks for the update. I re-reviewed the PDF image handling changes and I still see production blockers, so I am requesting changes again.

The max-size guard still runs after `Http::get()` has downloaded the full body, so chunked responses without `Content-Length` can consume memory/bandwidth before rejection. Please enforce the byte cap during streaming/download.

A remote URL with a `/forms/assets/...` path can still be treated as a local asset candidate and return a colliding local file. Please only resolve reserved asset paths locally when the expected app origin matches.

Also, the public-only DNS validation is not pinned into the actual cURL request, which leaves a DNS-rebinding/SSRF gap. Please ensure the request uses the validated public IP(s) rather than allowing a second unvalidated resolution.

JhumanJ (changes requested):
Thanks for the update. I re-reviewed the PDF image handling changes and I still see production blockers, so I am requesting changes again.

The max-size guard still runs after `Http::get()` has downloaded the full body, so chunked responses without `Content-Length` can consume memory/bandwidth before rejection. Please enforce the byte cap during streaming/download.

A remote URL with a `/forms/assets/...` path can still be treated as a local asset candidate and return a colliding local file. Please only resolve reserved asset paths locally when the expected app origin matches.

Also, the public-only DNS validation is not pinned into the actual cURL request, which leaves a DNS-rebinding/SSRF gap. Please ensure the request uses the validated public IP(s) rather than allowing a second unvalidated resolution.

JhumanJ (changes requested):
Thanks for the update. I re-reviewed the PDF image handling changes and I still see production blockers, so I am requesting changes again.

The max-size guard still runs after `Http::get()` has downloaded the full body, so chunked responses without `Content-Length` can consume memory/bandwidth before rejection. Please enforce the byte cap during streaming/download.

A remote URL with a `/forms/assets/...` path can still be treated as a local asset candidate and return a colliding local file. Please only resolve reserved asset paths locally when the expected app origin matches.

Also, the public-only DNS validation is not pinned into the actual cURL request, which leaves a DNS-rebinding/SSRF gap. Please ensure the request uses the validated public IP(s) rather than allowing a second unvalidated resolution.

chiragchhatrala:
Done

PR #1137, opened 2026-05-29
Enhance user identities migration logic and testing

chiragchhatrala:
- Updated the user identities migration to prevent dropping the table if it already exists, ensuring a clean migration state.
- Introduced a new test suite for the user identities migration, verifying that existing records are preserved and new records are created correctly when the migration is run.
- Added tests to confirm the correct handling of user identities in various scenarios, enhancing overall migration reliability.

### DIFF
--- a/api/database/migrations/2025_11_17_105205_create_user_identities_table.php
+++ b/api/database/migrations/2025_11_17_105205_create_user_identities_table.php
@@ -10,8 +10,9 @@ return new class () extends Migration {
      */
     public function up(): void
     {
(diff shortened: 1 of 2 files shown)


JhumanJ (approved):
Approve.

The migration change is small and coherent with the goal of preserving existing `user_identities`. The added tests cover the important cases: preserving an existing table and recreating it when missing.

I also ran the targeted Pest test successfully in an isolated worktree. A full PHPStan run failed on an existing unrelated issue in `app/Services/Tax/StripeExportDatasetService.php`, not on this PR.

PR #1138, opened 2026-05-29
Refactor TemplateController and FormTemplateRequest for improved attr…

chiragchhatrala:
…ibute handling

- Updated TemplateController to utilize getUpdateAttributes for template updates, enhancing attribute management.
- Enhanced FormTemplateRequest with new methods for retrieving mutable attributes and improved validation rules for publicly listed templates.
- Refactored logic to ensure consistent handling of attributes during template creation and updates, improving overall code clarity and maintainability.


JhumanJ (changes requested):
Changes requested.

Two issues need attention before approval:

1. Security / data integrity: `FormTemplateRequest.php` still lets the request body `id` drive the `unique:templates,slug,...` ignore id when there is no route id. On `POST /templates`, an authenticated user can pass `id=<existing template id>` with the same `slug` and bypass validation. Since there is no DB unique index on `templates.slug`, duplicate slugs can be persisted. Please only use the route id for updates, and no ignore id for creates.

2. Validation / runtime mismatch: fields like `types`, `industries`, and `related_templates` are validated as `nullable|array`, but `input("types", [])` preserves explicit `null`. This can write `null` into JSON columns that are expected/defaulted as arrays and can later break code expecting arrays. Please normalize explicit nulls to `[]` or make the validation non-nullable.

chiragchhatrala:
Done

api/app/Http/Requests/Templates/FormTemplateRequest.php
```diff
@@ -33,54 +33,96 @@ class FormTemplateRequest extends FormRequest
...
+     */
+    private function getMutableAttributes(): array
+    {
+        $attributes = [
+            'name' => $this->input('name'),
+            'slug' => $this->input('slug'),
+            'short_description' => $this->input('short_description'),
+            'description' => $this->input('description'),
+            'image_url' => $this->input('image_url'),
+            'structure' => $this->cleanFormStructure($this->input('form', [])),
```
JhumanJ, line 95: These optional list fields are always added to the update payload. If a client omits `types`/`industries`/`related_templates`/`questions`, `arrayInput()` converts the absence to `[]`, which clears existing metadata. Please only include these keys when present on update, or make the endpoint a full-replace contract with required fields.

JhumanJ (changes requested):
The previous safety concerns around null input and validation look improved. I am requesting changes for a remaining update regression that can silently clear template metadata.

The list fields are optional in the request rules, but `getUpdateAttributes()` still includes `types`, `industries`, `related_templates`, and `questions` unconditionally through `getMutableAttributes()`. When an update request omits one of those fields, `arrayInput()` converts the absence to `[]`, so an otherwise partial update clears existing values. Please preserve omitted fields on update, or make the endpoint a full-replace contract and require those arrays. A regression test for an update omitting these fields would be useful here.

chiragchhatrala:
Done

JhumanJ (approved):
Re-reviewed after the latest update. The previous issue is fixed: partial template updates no longer clear omitted array fields, and the regression test covers preserving `types`, `industries`, `related_templates`, and `questions` when omitted. Targeted local Pest test passed. Approved.

PR #1140, opened 2026-05-29
Enhance JWT validation configuration and testing

chiragchhatrala:
- Added `JWT_SKIP_IP_UA_VALIDATION` environment variable to `docker-compose.dev.yml` for local development, allowing JWT User Agent validation to be skipped.
- Updated `.env.docker` to set `JWT_SKIP_IP_UA_VALIDATION` to `false` for production environments.
- Introduced tests in `AuthenticateJWTTest.php` to verify the behavior of JWT validation based on the new configuration, ensuring proper handling of user agent checks.
- Improved documentation to clarify the purpose and usage of `JWT_SKIP_IP_UA_VALIDATION` in different environments.


JhumanJ (changes requested):
Changes requested.

Two points should be fixed before approval:

1. `scripts/docker-setup.sh` only detects lowercase unquoted truthy values for `JWT_SKIP_IP_UA_VALIDATION`. Values such as `"true"`, `(true)`, `TRUE`, or `true # comment` can still disable Laravel validation without being blocked by the Docker production guard. Please normalize or parse the value more robustly.

2. `AuthenticateJWTTest.php` does not actually prove the skip short-circuit. The request has no token, so even without `config("app.jwt_skip_ip_ua_validation")`, the middleware would fall into the `JWTException` catch path and call `$next`. Please make the test assert that token parsing is skipped, or provide an invalid UA/IP payload and verify that the request passes only when the flag is true.

chiragchhatrala:
Done

JhumanJ (approved):
Looks good. The previous JWT skip-IP/UA concerns appear addressed: prod defaults remain safe, Docker/dev opts in intentionally, setup normalizes truthy values, and tests assert JWT parsing is bypassed when configured. CI is green.

PR #1141, opened 2026-05-29
ESC-857 = Enhance form submission handling by adding rich text field support an…

chiragchhatrala:
…d improving email notification tests.

- Updated `FormSubmissionFormatter` to clean rich text field values using the Purify library, ensuring safe HTML output.
- Introduced new test cases in `EmailNotificationTest` to validate rendering of rich text fields in email content, enhancing coverage for hidden fields and rich text submissions.

These changes improve data integrity and user experience in form submissions and notifications.

### DIFF
--- a/api/app/Service/Forms/FormSubmissionFormatter.php
+++ b/api/app/Service/Forms/FormSubmissionFormatter.php
@@ -5,8 +5,9 @@ namespace App\Service\Forms;
 use App\Models\Forms\Form;
 use App\Service\Storage\FilenameUrlEncoder;
 use Carbon\Carbon;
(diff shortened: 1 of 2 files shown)


JhumanJ (approved):
Approve.

I reviewed the rich text submission formatting path from formatter to notification email rendering. Rendering rich text as HTML only after `Purify::clean()` is consistent with the Blade template using raw output only when `value_is_html` is set.

Non-blocking follow-up: since this enables raw HTML rendering in email for rich text fields, a direct assertion that hostile HTML such as scripts, event handlers, or javascript URLs is stripped would make the coverage stronger.

PR #1142, opened 2026-05-29
ESC-883 = Enhance API rate limiting for export endpoints

chiragchhatrala:
- Introduced dedicated rate limits for CSV export and export status endpoints to prevent long-running exports from being blocked by general API limits.
- Updated route definitions to apply the new throttling middleware for export operations, ensuring smoother user experience during data exports.
- Added tests to verify the functionality of the new rate limits and ensure compliance with expected behavior.

These changes improve the performance and reliability of the export features in the application.

### DIFF
--- a/api/app/Providers/RouteServiceProvider.php
+++ b/api/app/Providers/RouteServiceProvider.php
@@ -56,6 +56,16 @@ class RouteServiceProvider extends ServiceProvider
             return Limit::perMinute(30)->by($request->user()?->id ?: $request->ip());
         });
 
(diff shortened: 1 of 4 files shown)


JhumanJ (approved):
Approve.

I reviewed the export throttling changes. The dedicated `export` and `export-status` rate limiters look coherent, and removing the global API throttle from these export routes avoids blocking status polling after unrelated API traffic.

Non-blocking follow-up: the added test covers status polling after the global throttle is exhausted. It would be useful to also explicitly test that `POST /export` escapes the global throttle and is limited by its own `10/min` bucket.

PR #1143, opened 2026-06-02
BE-001 - Refactor Zapier integration authorization and enhance testing for rea…

chiragchhatrala:
…donly workspace members

- Updated authorization checks in `IntegrationController` to use 'manageIntegrations' instead of 'view', ensuring proper permissions for integration management.
- Enhanced tests in `IntegrationsTest.php` to verify that readonly workspace members cannot create, delete, or poll submissions for integrations, even with integration token abilities.
- Introduced a new function for managing integration token abilities, improving test coverage and clarity.

These changes strengthen the security and functionality of the Zapier integration features.


api/app/Http/Controllers/Integrations/Zapier/IntegrationController.php
```diff
@@ -18,7 +18,7 @@ public function store(CreateIntegrationRequest $request)
     {
         $form = $request->form();
 
-        $this->authorize('view', $form);
+        $this->authorize('manageIntegrations', $form);
```
JhumanJ, line 21: This policy currently goes through `canPerformWriteOperation()`, so Sanctum tokens need `forms-write` in addition to `manage-integrations`. That appears stricter than the external Zapier route contract and may break existing integration tokens.

api/tests/Feature/Zapier/IntegrationsTest.php
```diff
@@ -10,6 +10,11 @@
 use function Pest\Laravel\post;
 use function PHPUnit\Framework\assertEquals;
 
+function zapierIntegrationTokenAbilities(): array
```
JhumanJ, line 13: This helper gives every happy-path token `forms-write`, which hides the compatibility regression for existing Zapier tokens that only have `manage-integrations`. Please keep a test proving `manage-integrations` alone still works for writable workspace members, or intentionally update the API contract/docs.

JhumanJ (changes requested):
I agree with the intent to prevent read-only users from managing Zapier integrations, but this implementation appears to over-tighten the external token contract.

`FormPolicy::manageIntegrations()` goes through `canPerformWriteOperation()`, which means Sanctum/API tokens now need `forms-write` in addition to `manage-integrations`. Existing Zapier/integration tokens that only have `manage-integrations` can regress, and the route-level ability contract still advertises `manage-integrations` as the relevant ability. Please either keep `manage-integrations` sufficient for writable workspace members, or intentionally migrate/document the stricter token requirement and update route/API docs. A test for a writable member token with only `manage-integrations` would make the intended contract clear.

chiragchhatrala:
Done

JhumanJ (approved):

PR #1146, opened 2026-06-03
BE-004 - Enhance form submission handling and validation for special characters

chiragchhatrala:
- Updated `StoreFormSubmissionJob` to allow completed submissions to be updated when the `allowCompletedUpdate` flag is set, improving flexibility in submission management.
- Enhanced `FormSubmissionController` to utilize the new job configuration for handling submissions.
- Introduced validation logic for select options to correctly process special characters, ensuring accurate data handling during form submissions.
- Added comprehensive tests to validate the handling of special characters in form submissions, enhancing overall data integrity and user experience.

These changes improve the robustness of form submission handling and ensure compliance with user input requirements.

### DIFF
--- a/api/app/Http/Controllers/Forms/FormSubmissionController.php
+++ b/api/app/Http/Controllers/Forms/FormSubmissionController.php
@@ -86,7 +86,9 @@ class FormSubmissionController extends Controller
 
         $submissionData = $request->validated();
         $submissionData['submission_id'] = $submission->id;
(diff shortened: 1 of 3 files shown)


JhumanJ (approved):
Approved. This protects completed submissions from late partial autosaves while preserving admin edit behavior. CI is green. Non-blocking follow-up: the terminal guard still runs after `getFormData()`, so file/signature processing may happen before the late partial is skipped; consider moving that guard earlier to avoid orphaned uploads.

PR #1147, opened 2026-06-03
BE-005 - Implement auto-increment ID generation for forms and enhance form sub…

chiragchhatrala:
…mission handling

- Introduced `FormAutoIncrementSequence` service to manage atomic allocation of auto-increment IDs for forms.
- Updated `StoreFormSubmissionJob` to utilize the new service for generating IDs, improving submission logic and user experience.
- Added migration to include `auto_increment_sequence` in the forms table, ensuring proper database structure for ID management.
- Developed comprehensive tests for the new ID generation logic, ensuring reliability and correctness in various scenarios.

These changes enhance the functionality of form submissions by providing a robust mechanism for auto-increment ID generation, improving overall data integrity and user experience.


api/app/Service/Forms/FormAutoIncrementSequence.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+        return (string) DB::transaction(function () use ($form) {
+            $lockedForm = Form::query()
+                ->whereKey($form->id)
+                ->lockForUpdate()
+                ->firstOrFail();
+
+            $next = $lockedForm->auto_increment_sequence + 1;
+            $lockedForm->auto_increment_sequence = $next;
+            $lockedForm->save();
```
JhumanJ: Saving the Eloquent model here will dispatch normal `Form` saved events for every ID allocation. That can trigger form-spam checks/versioning for each submission. Please update this counter quietly, e.g. with a query-builder update, `saveQuietly()`, or `withoutEvents()` inside the lock.

api/app/Jobs/Form/StoreFormSubmissionJob.php
```diff
@@ -449,4 +451,17 @@ private function workspaceHasIdGenerationAccess(): bool
...
         return $workspace->hasFeature(Feature::ID_GENERATION);
     }
+
+    private function autoIncrementIdValue(): string
+    {
+        if (!$this->workspaceHasIdGenerationAccess()) {
+            return 'Please upgrade your OpenForm subscription to use our ID generation features';
+        }
+
+        if ($this->allocatedAutoIncrementId === null) {
```
JhumanJ: This allocates the sequence during data preparation. At this point we do not yet know whether the job will create a final submission, update an existing one, or skip a late partial, so IDs can be consumed without a completed submission.

JhumanJ (changes requested):
The atomic sequence approach is the right direction for the original race condition, but I am requesting changes before merge because this implementation has two important side effects:

1. `FormAutoIncrementSequence::allocateNext()` saves the `Form` model for every allocation. That dispatches the normal form saved events, which can trigger form-spam checks/versioning or other form-level side effects on each submission. The counter update should be done quietly inside the row lock.
2. The sequence is allocated while preparing submission data, before the job knows whether it will create a new completed submission, update an existing one, or skip a late partial. That can consume IDs without a final submission. If gaps are acceptable, the product/ops behavior should be explicit and tested; if not, allocation needs to happen inside the final storage path only for new completed submissions, preserving existing IDs on edits/partials.

chiragchhatrala:
Done

JhumanJ (approved):
Re-reviewed locally after the latest updates. The previous requested changes are resolved and the current checks are green. Approving for merge.

PR #1150, opened 2026-06-04
BE-017 - Refactor migration to correct access token removal from user_workspac…

chiragchhatrala:
…e table

- Updated the migration file to change the target table from 'workspaces' to 'user_workspace' for the removal of the access token column.
- Ensured that the migration correctly reflects the intended schema changes for user workspaces, enhancing data integrity and consistency.

These changes improve the accuracy of the database migration process related to user workspaces.


JhumanJ (approved):
Looks good. The rollback now targets the same `user_workspace` table modified by the migration's `up()`, restoring `access_token`/`is_owner` and dropping `role` in the right place. CI is green.

PR #1151, opened 2026-06-05
Refactor workspace user role management and feature access logic

chiragchhatrala:
- Updated `WorkspaceUserController` to change authorization from 'adminAction' to 'inviteUser', aligning with new role management requirements.
- Enhanced `RemoveWorkspaceGuests` job to check for 'invite_user' feature instead of 'workspaces.multiple', ensuring correct access control for workspace guests.
- Modified `WorkspacePolicy` to reflect the updated requirements for inviting users, specifying that a Business plan is necessary for role-based invitations.
- Introduced a new method in `PlanAccessService` to check for license feature grants, improving feature access validation.
- Updated configuration files to reflect the new feature mappings and licensing structure.
- Added migration to rename 'multi_user.roles' to 'invite_user' in plan overrides, ensuring consistency across the application.
- Enhanced tests to cover new role management and feature access scenarios, ensuring robust validation of user permissions and workspace features.

These changes improve the clarity and maintainability of user role management and feature access across the application.


api/app/Http/Controllers/Auth/UserInviteController.php
```diff
@@ -41,7 +41,7 @@ public function resendInvite(Workspace $workspace, $inviteId)
 
     public function cancelInvite(Workspace $workspace, $inviteId)
     {
-        $this->authorize('adminAction', $workspace);
+        $this->authorize('inviteUser', $workspace);
```
JhumanJ: This uses `inviteUser` for cancellation, but canceling an existing invite does not create a new seat. Because `inviteUser` also enforces Business/AppSumo seat rules, an over-limit workspace may be unable to cancel stale invites. Please use a management permission here that does not depend on the ability to add a user.

api/app/Http/Controllers/WorkspaceUserController.php
```diff
@@ -91,7 +91,7 @@ private function inviteUser(Workspace $workspace, string $email, string $role)
 
     public function updateUserRole(Request $request, Workspace $workspace, User $user)
     {
-        $this->authorize('adminAction', $workspace);
+        $this->authorize('inviteUser', $workspace);
```
JhumanJ: Updating an existing member's role is not the same operation as inviting a new user. Reusing `inviteUser` makes role management depend on plan/seat checks intended for adding users; please split this permission from the add-user path.

JhumanJ (changes requested):
This is a good direction for making the `invite_user` feature explicit, but I am requesting changes because the new policy is being reused for actions that are not actually adding a new seat.

`inviteUser` now applies both the Business-plan feature gate and AppSumo seat limits. The PR uses it for updating roles, resending invites, and canceling invites. Those operations manage existing users/invites; they should not depend on the workspace being allowed to add another user. For example, an AppSumo workspace already at its user limit could be blocked from canceling a stale invite or changing an existing member role. Please split "can add/invite a new user" from "can manage existing members/invites" and update the UI/tests so admins without the add-user feature do not just see actions that fail with 403.

chiragchhatrala:
Done

JhumanJ (changes requested):
Thanks for the update. I re-reviewed this after the latest changes, and the previous concern is still only partially addressed.

`manageWorkspaceMembers` still requires the `invite_user` feature and is still used for existing-member operations such as role updates, resend invite, and cancel invite. That means an admin can still be blocked from managing existing users/invites even though those actions are not adding a new seat. Please keep the add-seat checks isolated to the invite/new-user path, and let existing-member management rely on admin/write authorization.

I also found a related mixed-owner edge case: AppSumo can grant `invite_user`, but the seat-limit check is based on the first billing owner, which can behave incorrectly for mixed Pro/AppSumo workspaces. Please make the limit source-aware and cover the mixed-owner case in tests.

JhumanJ (changes requested):
Thanks for the update. I re-reviewed this after the latest changes, and the previous concern is still only partially addressed.

`manageWorkspaceMembers` still requires the `invite_user` feature and is still used for existing-member operations such as role updates, resend invite, and cancel invite. That means an admin can still be blocked from managing existing users/invites even though those actions are not adding a new seat. Please keep the add-seat checks isolated to the invite/new-user path, and let existing-member management rely on admin/write authorization.

I also found a related mixed-owner edge case: AppSumo can grant `invite_user`, but the seat-limit check is based on the first billing owner, which can behave incorrectly for mixed Pro/AppSumo workspaces. Please make the limit source-aware and cover the mixed-owner case in tests.

JhumanJ (changes requested):
Thanks for the update. I re-reviewed this after the latest changes, and the previous concern is still only partially addressed.

`manageWorkspaceMembers` still requires the `invite_user` feature and is still used for existing-member operations such as role updates, resend invite, and cancel invite. That means an admin can still be blocked from managing existing users/invites even though those actions are not adding a new seat. Please keep the add-seat checks isolated to the invite/new-user path, and let existing-member management rely on admin/write authorization.

I also found a related mixed-owner edge case: AppSumo can grant `invite_user`, but the seat-limit check is based on the first billing owner, which can behave incorrectly for mixed Pro/AppSumo workspaces. Please make the limit source-aware and cover the mixed-owner case in tests.

chiragchhatrala:
Done

JhumanJ:
Closing this PR because it gates user invitations behind Business, but our public pricing currently says Free/Pro include multi-user access with all members as admins, while Business unlocks roles & permissions.

We should handle this in a new, narrower PR that aligns the product with the public pricing page:
- keep invitations available for Free/Pro as admin-only collaboration;
- gate non-admin roles (`user`, `readonly`) behind Business+;
- update backend, frontend, and tests together;
- treat AppSumo / special licenses separately.

So this PR is no longer the right shape to merge as-is.

PR #1154, opened 2026-06-07
Adding inputMode for improved mobile phone UX

doubledods:
This change improves the mobile phone experience for end users. Their phone keyboard will pop up with the appropriate settings for the tel, email, and url inputs. For example, on email fields, mobile phone keyboards will show the @ symbol on the first page of the keyboard.

Using inputMode instead of nativeType, per the suggestions from PR 1119.


JhumanJ:
Thanks a lot for the contribution. I am going to close this one for now because the current change does not actually reach the rendered input element.

The PR adds `inputMode` from `BlockRenderer`, but `TextInput` does not currently expose an `inputMode` prop or bind it onto the inner `<input>`. With Vue fallthrough, the attribute lands on the wrapper instead, so the mobile keyboard behavior would not change for users.

A good follow-up would be to add `inputMode` support directly in `client/components/forms/core/TextInput.vue` and pass it through to the actual input. Happy to review again if you reopen with that adjustment.

Issue #1169, opened 2026-06-18
[BUG]  Negative Logic operators on Forms (`does_not_equal`, etc.) evaluate true against unanswered fields, causing premature hiding and broken positive+negative AND with or without groups


WayOfThKeyboard:
### Environment
- OpnForm **self-hosted**, version **1.12.1** (`jhumanj/opnform-api:1.12.1` / `jhumanj/opnform-client:1.12.1`)
- Reproduced in Firefox/Chromium; not browser-specific (logic runs both SSR and client-side)

### Summary
Negative comparators (`does_not_equal`, `does_not_contain`, `does_not_match_regex`, `content_length_does_not_equal`) treat an **unanswered/empty field as satisfying the condition**. An empty value "does not equal X" evaluates true, so a Hide Block action fires before the referenced field has been answered. This breaks any AND group that mixes a positive and a negative condition.

---

### Root cause

In `client/lib/forms/FormLogicConditionChecker.js`, negative comparators are implemented as the bare inverse of their positive counterpart, with no check that the field has actually been answered:

```js
case "does_not_equal":
  return !checkEquals(propertyCondition, value)
```

When `value` is empty/`undefined`, `checkEquals` returns `false`, so `does_not_equal` returns **`true`** for an unanswered field. The same pattern applies to `does_not_contain`, `does_not_match_regex`, and `content_length_does_not_equal`, across the text, number, select, multi_select, and matrix branches.

---

### Reproduction (positive + negative AND group)

Two-condition AND group on a multi-page form, with the two referenced fields on different pages:

| Conditions | Result |
|---|---|
| both **positive** (equals + equals) | works as expected |
| both **negative** (does_not_equal + does_not_equal) | works as expected |
| **positive + negative** (equals + does_not_equal) | **fails** — block hides prematurely |

When the positive condition's field is answered it becomes true, while the negative condition is *already* true against the still-empty second field, so the AND is satisfied before that field is reached and the block hides early.

**Minimal repro:** a Hide Block action with a single `does_not_equal` condition hides the block immediately — even before the form is filled out — because the empty field value satisfies the negative condition.

---

### Suggested fix
Guard the value-comparison negative operators so an unanswered field satisfies **no** value comparison (positive or negative). Leave `is_empty` / `is_not_empty` / `is_checked` / `is_not_checked` untouched — those legitimately test the empty state.

### Workaround
[...]

PR #1170, opened 2026-06-18
[codex] Fix negative logic on unanswered fields

JhumanJ:
## Summary

Fixes form logic negative value comparisons so unanswered fields no longer satisfy `does_not_equal`, `does_not_contain`, `does_not_match_regex`, or equivalent select/multi-select/matrix comparisons.

Refs #1169.

## Root cause

The frontend and backend condition checkers implemented negative value operators as bare inverses of their positive checks. When a referenced field had not been answered yet, the positive check returned false, so the negative check returned true and could prematurely satisfy mixed AND groups.

## Changes

- Add an answered-value guard before negative value comparisons in the Nuxt form logic checker.
- Mirror the same guard in the Laravel form logic checker.
- Add frontend and backend regression tests for unanswered negative operators and mixed positive/negative AND groups.
- Update the existing backend matrix-empty expectation so empty matrix values do not match value comparisons before an answer exists.

## Validation

- `npm run test -- --run test/unit/FormLogicConditionChecker.test.ts`
- `npm run lint`
- `php vendor/bin/pest tests/Unit/Service/Forms/FormLogicConditionCheckerTest.php`
- `php vendor/bin/pint --test app/Service/Forms/FormLogicConditionChecker.php tests/Unit/Service/Forms/FormLogicConditionCheckerTest.php`

Note: backend Pest run passes with existing warning/deprecation output unrelated to this change.

### DIFF
--- a/api/app/Service/Forms/FormLogicConditionChecker.php
+++ b/api/app/Service/Forms/FormLogicConditionChecker.php
@@ -354,6 +354,23 @@ class FormLogicConditionChecker
             is_numeric($fieldValue);
     }
 
+    private function hasAnswerValue($fieldValue): bool
+    {
+        if ($fieldValue === null) {
+            return false;
+        }
+
+        if (is_string($fieldValue)) {
+            return $fieldValue !== '';
+        }
+
+        if (is_array($fieldValue)) {
+            return count($fieldValue) > 0;
+        }
+
+        return true;
+    }
+
     private function checkGreaterThan($condition, $fieldValue): bool
     {
         if (!$this->areValidNumbers($condition, $fieldValue)) {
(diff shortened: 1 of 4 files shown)


JhumanJ:
Closing this draft after re-evaluating the product semantics.

The current behavior is defensible: an unanswered/empty value can be treated as "not equal to X" unless the form author explicitly requires an answer. For the stricter behavior, builders can already express it with an additional guard such as:

`field is_not_empty AND field does_not_equal X`

Changing `does_not_equal` / `does_not_contain` to return false for unanswered fields would be a semantic change and could break existing forms that rely on the current behavior. Since this is more of a product decision than a clear correctness bug, we should not merge this PR as-is.

PR #1183, opened 2026-06-30
Add MCP server tools and resources for form management

chiragchhatrala:
- Introduced `FieldTypesResource` to provide a catalog of available form field types.
- Added `OpnFormGuestServer` and `OpnFormServer` to handle guest and authenticated form operations.
- Implemented various tools for form management including `CreateFormTool`, `DeleteFormTool`, `DuplicateFormTool`, `GetFormTool`, `ListFormsTool`, `UpdateFormTool`, and `DraftFormTool`.
- Created submission handling tools: `GetSubmissionTool` and `ListSubmissionsTool`.
- Added workspace management tool `ListWorkspacesTool`.
- Updated `composer.json` to include `laravel/mcp` dependency and modified `composer.lock` accordingly.
- Added routes for MCP servers and created feature tests for form tools.

These changes enhance the functionality and usability of the MCP for form-related operations, allowing for comprehensive management of forms and submissions.


api/app/Mcp/Tools/Forms/CreateFormTool.php
```diff
@@ -0,0 +1,151 @@
...
+            if (empty($field['id'])) {
+                $field['id'] = Str::uuid()->toString();
+            }
+
+            return $field;
+        })->values()->all();
+
+        $user = $request->user();
+
+        if ($user && !empty($validated['workspace_id'])) {
```
JhumanJ, line 39: [P1] This makes creation persist immediately whenever the agent has a token and `workspace_id`. For agent-generated new forms, I think this should stay draft-first: create a draft/preview/handoff artifact and let the user finalize inside OpnForm. Real authenticated mutation tools are still useful for existing forms, but creation is much easier to do accidentally and should not bypass the editor flow.

api/app/Mcp/Tools/Forms/CreateFormTool.php
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    private function persistForm(Request $request, $user, array $validated, array $properties): ResponseFactory
+    {
+        $workspace = Workspace::findOrFail($validated['workspace_id']);
+
+        Gate::forUser($user)->authorize('ownsWorkspace', $workspace);
+        Gate::forUser($user)->authorize('create', [Form::class, $workspace]);
+
+        $formData = collect($request->all())
```
JhumanJ: [P1] This direct `request->all()` -> `Form::create()` path bypasses the normal form write invariants (`StoreFormRequest`/`UserFormRequest`, `FormPropertiesRule`, and `FormCleaner`). That lets MCP persist field structures and feature flags the editor/API would normally normalize, reject, or clean. Please move real create/update writes behind a shared form mutation service or otherwise reuse the same validation/cleaning chain as `FormController`.

api/app/Mcp/Tools/Forms/UpdateFormTool.php
```diff
@@ -0,0 +1,103 @@
...
+        $updateData = collect($request->all())
+            ->except(['form_id'])
+            ->only(self::ALLOWED_FIELDS)
+            ->all();
+
+        if (empty($updateData)) {
+            return Response::error('No valid fields provided to update. Provide at least one field like title, visibility, or properties.');
+        }
+
+        if (isset($updateData['properties'])) {
```
JhumanJ, line 71: [P1] `properties` replacement needs the same normalization as create. Right now new fields sent by the agent can omit `id`, but the removed-properties calculation assumes every field has one and the update then persists the raw array. Please backfill stable IDs for new fields and validate the resulting `properties` through the same `FormPropertiesRule` path before saving.

api/app/Mcp/Servers/OpnFormServer.php
```diff
@@ -0,0 +1,42 @@
...
+use App\Mcp\Tools\Submissions\ListSubmissionsTool;
+use App\Mcp\Tools\Workspaces\ListWorkspacesTool;
+use Laravel\Mcp\Server;
+use Laravel\Mcp\Server\Attributes\Instructions;
+use Laravel\Mcp\Server\Attributes\Name;
+use Laravel\Mcp\Server\Attributes\Version;
+
+#[Name('OpnForm')]
+#[Version('1.0.0')]
+#[Instructions('OpnForm MCP Server. Manage forms, view submissions, and build no-code forms programmatically. Without a Bearer token only draft-form and field-types are available. With a token, use list-workspaces first to get a workspace_id, then list-forms or create-form.')]
```
JhumanJ: [P2] The instructions mention `draft-form`, but this server does not register a `draft-form` tool. Agents will try to call a non-existent tool. Either change this to `create-form` draft mode or split out a real draft creation tool and register it.

JhumanJ (changes requested):
Thanks for pushing this. I like the direction of exposing real authenticated MCP tools for existing forms, but I do not think this is ready to merge yet. The main issue is that the PR mixes draft-first creation with direct writes to production forms, while the real write path bypasses the form editor/API invariants.

Requested changes:

1. Split creation from real mutations: creation from an agent should produce a draft/preview/handoff flow first, rather than immediately saving a live Form whenever a token and workspace_id are present. Authenticated tools for existing forms are still valuable, but should be a separate contract.
2. Route all real form create/update operations through the same validation, normalization, and cleaning used by the normal editor/API path (`UserFormRequest`, `FormPropertiesRule`, `FormCleaner`, removed_properties handling, feature-tier cleanup).
3. Make field updates safer: at minimum backfill IDs on new properties during update and validate the full property payload; ideally add a patch-style form editing tool instead of requiring agents to replace the whole `properties` array.
4. Fix the stale `draft-form` server instruction and the `git diff --check` trailing blank line in `docs/mcp/tools.mdx`.

JhumanJ:
One additional architecture note: I would avoid ending up with two separate MCP servers/endpoints, one for guest/draft flows and one for authenticated account operations. The cleaner shape is a single OpnForm MCP server with optional auth and clear capability boundaries:

- unauthenticated: field/schema resources + draft creation/preview/handoff only;
- authenticated: the same draft flow, plus real tools for existing resources such as list/get/update/delete/duplicate forms and submissions;
- real writes should still go through the shared validation/cleaning/mutation path.

That keeps the client integration simple (`/mcp` is always the OpnForm MCP server), while the server decides which tools/actions are available from the request context. The split should be at the tool/permission layer, not at the server layer.

chiragchhatrala:
Done

JhumanJ:
Implemented the requested fixes in `a4b0132a`:

- `create-form` now always returns draft JSON and never persists a live form, even with authentication + `workspace_id`.
- `update-form` still updates existing forms for real, using `FormCleaner` with the existing form context so custom-domain forms keep `nf-code` blocks.
- Removed the extra MCP cleaning path that called `Purify` without importing it.
- Updated MCP server instructions, docs, and client copy so the guest/auth behavior is described as one server + one `create-form` draft flow.
- Added regression coverage for draft-only creation, text sanitization on update, and preserving `nf-code` for custom-domain forms.

Local validation:

- `php -l` on changed PHP files
- `./vendor/bin/pest tests/Feature/Mcp tests/Unit/Service/FormCleanerTest.php --display-warnings` — 53 passed, 171 assertions
- `npm run lint`
- `PATH="/Users/jhumanj/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH" npm run test -- --run` — 41 files passed, 644 tests
- `git diff --check origin/main...HEAD`

GitHub checks observed after push: code linters, shared-file sync, Mintlify, client linters/tests/build, and PHP 8.2/8.3 mysql/pgsql are passing. The downstream `e2e` job was still pending when I stopped watching.

JhumanJ:
Superseded by the reviewed, validated MCP replacement stack: #1250, #1251, #1252, #1253, #1254, #1255, and #1256. The new stack covers the unified guest/authenticated server, secure draft preview and claim flow, OAuth, safe form and submission management, and the Agent Plugins package.

PR #1187, opened 2026-07-05
fix: normalize public form favicon scheme

ahfoysal:
## Summary
- normalize same-host public form favicon URLs from `http://` to `https://` when the form page is served over HTTPS
- keep local HTTP/self-hosted pages and external favicon URLs unchanged
- add unit coverage for favicon URL normalization

## Verification
- `npm install`
- `npm test -- --run test/unit/public-page-meta.test.ts`
- `npx eslint "pages/forms/[slug]/index.vue" lib/forms/public-page-meta.js`
- `npm run build`
- `git diff --check`

Closes #1068.

### DIFF
--- a/client/lib/forms/public-page-meta.js
+++ b/client/lib/forms/public-page-meta.js
@@ -1,0 +1,17 @@
+export function normalizePublicPageAssetUrl (assetUrl, pageUrl) {
+  if (!assetUrl || !pageUrl) return assetUrl
+
+  try {
+    const page = new URL(pageUrl)
+    const asset = new URL(assetUrl, page)
+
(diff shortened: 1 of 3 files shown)


JhumanJ:
Thanks for the PR.

I am closing this because the issue is not in the public form rendering code. The favicon URL is already stored as `http://...` when the asset is uploaded.

This is most likely caused by the self-hosted instance configuration, especially `APP_URL` being set to `http://...` instead of `https://...`.

We should fix the instance URL/proxy configuration so the backend generates the correct public asset URL from the start, rather than normalizing the favicon URL in the frontend after the bad value has already been stored.

PR #1216, opened 2026-07-22
feat: renormalize source_page indices in PDF template updates

chiragchhatrala:
- Implemented logic to renormalize source_page values to sequential indices after page removal from the manifest, ensuring correct page exports and re-edits.
- Added comprehensive tests to verify the renormalization process under various scenarios, including mixed source and blank pages, and multiple saves without errors.

### DIFF
--- a/api/app/Http/Controllers/Pdf/PdfTemplateController.php
+++ b/api/app/Http/Controllers/Pdf/PdfTemplateController.php
@@ -140,13 +140,13 @@ class PdfTemplateController extends Controller
                 ], 422);
             }
 
-            // Renormalize source_page to sequential 1-based indices matching
-            // the rebuilt file. Without this, stale source_page values (from the
-            // original upload) cause wrong/blank pages on export and re-edit.
-            $sequentialPage = 1;
-            foreach ($pageManifest as &$entry) {
+            // Renormalize source_page to match the physical position in the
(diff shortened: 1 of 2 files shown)


JhumanJ:
**[P1] Blank pages offset the physical `source_page` indices**

`PdfTemplateRebuildService` writes one physical PDF page for every manifest entry, including blank pages. However, the new renormalization counter advances only for `source` entries. For a manifest such as `[source page 1, blank, source page 4]`, the rebuilt file contains three physical pages, but the stored manifest becomes `[1, null, 2]` instead of `[1, null, 3]`.

This makes the source page after the blank resolve to the blank physical page during export or re-edit. A subsequent save then permanently rebuilds the stored template with the original source page missing. I reproduced this with distinct page sizes: the first save produced `120, 120, 180`, while export/second save using the renormalized manifest produced `120, 120, 120`.

Please renormalize source entries to their physical manifest position (`$index + 1`) while keeping blank entries at `null`. The mixed-page test should expect `3`, and it would be useful to add a second-save/export regression test that verifies distinguishable page content or dimensions after an intermediate blank page.

**Verdict: not safe to merge until this case is fixed.**

chiragchhatrala:
Fixed

JhumanJ (approved):

PR #1240, opened 2026-08-10
Add an optional in-product copilot for form creators (off by default)Holostaff copilot

xxl007:
Hey OpnForm maintainers 👋, I'm Galem, adding an optional in-product copilot for form creators. I have added some screen captures below, so you can visualize what I mean by this.

05-live-intervention

That is someone using OpnForm for the first time. They open the form builder, reach the style modal, and stop. A little later, help arrives (the copilot). It learned OpnForm by reading this repo.

It can also act when asked 👇👀:

06-voice-and-act

Here the user asked by voice for a phone number field. The copilot opened the Add Block and added one, naming each step before taking it.

The copilot is off by default.

## How did I add this?

I used the Holostaff library (which I build 😎), ran a command through its CLI in a checkout of this repo. It scanned the code and drew OpnForm's customer journey: the workflows, the places where a creator gets stuck, and what a helpful voice should say at each one. That map is where the copilot's knowledge comes from, so it arrives fully knowing OpnForm.

03-map-create-a-form

I wrote the full map up, workflow by workflow, here:
[We scanned OpnForm](https://www.holostaff.ai/blog/we-scanned-opnform).

One example from the map. The scan noticed that a guest can build a whole form without an account, and is only invited to sign up when they hit Save. A guest who drifts away before saving is exactly who the copilot is for: it notices someone who has been building for a while without saving, and invites them to keep their work while the form is still on their screen. It would say something like: "Looking good! Create a free account to save your form and start collecting responses."

It picks its moments in the other direction too. On the submissions export the copilot stays quiet, because your tooltip already covers it. A helper that talks all the time is worse than none.

Two things that matter for a project whose pitch is self-hosting:

- The copilot never runs inside iframes or on public form pages. It is for the person building the form, never the person answering it.
- Before the scan uploads anything to the copilot's context, it shows exactly what leaves the machine. Source code, file contents, .env and git history stay put.

02-trust-report

## What merging does

The diff is 5 files in `client/`, modelled on your `crisp.client.js`: one plugin that returns early unless two env vars are set. I measured what "off" means on a build of this branch, instead of asserting it:

| | ids blank | ids set |
|---|---|---|
| requests to Holostaff | **0** | 6 |
| copilot in the DOM | **no** | yes |
| SDK chunk (22 KB gzip) | idle prefetch only | fetched |
| session recorder (rrweb, 80 KB gzip) | not fetched | fetched |
| voice (livekit-client) | not fetched | not fetched until a conversation opens |

The one real cost when off: Nuxt emits an idle `prefetch` hint for the 22 KB SDK chunk. It never executes and nothing is sent. If you want that hint gone too, I will strip it from the build manifest in this PR.

To be complete, the table is browser bytes only. The two dependencies land in `client/package.json` for every contributor and self-hoster either way: install time, CI time, supply-chain surface. `@holostaff/sdk ^0.10.9` is new with a small user base, `livekit-client ^2.21.0` is the voice transport, and rrweb resolves to 2.1.1 stable in the lockfile. The SDK is Apache-2.0, which works with AGPLv3. The CLI is open source; the service behind it is not.

## How this would roll out

Merging turns nothing on for opnform.com. It gives anyone self-hosting OpnForm two env vars they can choose to fill in, on their own install. Setup is one command, documented at [docs.holostaff.ai/cli/presets (https://docs.holostaff.ai/cli/presets.html), and I am happy to move that write-up into your docs instead, so you control what your users read.

And if a vendor plugin inside core is more than you want to take on, closing this costs your self-hosters nothing: the same plugin can be added to any checkout with our CLI, and I keep that path working and tested. Merging just saves each of them from carrying a patch.

Happy to answer anything here, and to adapt the approach to what works for OpnForm. You can also reach me at galem@holostaff.ai.


JhumanJ:
Thanks for doing this, but we're about to release our own MCP and built-in chat so we won't need this for now.

xxl007:
Hey @JhumanJ, congrats on the MCP and built-in chat, sounds like a great direction.

One clarification, because the PR clearly didn't get this across: this is neither of those. A chat waits for the user to ask. Holostaff watches the session and steps in on its own when someone looks stuck (like the user above who builds a whole form and drifts away before saving). Such users may leave without initiating a chat. And there is no MCP underneath: it works the browser directly, seeing the same screen as the user and clicking the same buttons.

So I see it as a third option next to chat and MCP, not a replacement for either. If it becomes interesting for the OpnForm community once yours ship, happy to pick this up again. Meanwhile anyone self-hosting who wants the proactive kind can add it to their own checkout with our CLI, and I keep that path working and tested.

Good luck with the release!

PR #1242, opened 2026-08-11
Opnform oAuth

chiragchhatrala:


JhumanJ:
Superseded by the reviewed, validated MCP replacement stack: #1250, #1251, #1252, #1253, #1254, #1255, and #1256. OAuth is now integrated as optional account access on the same MCP endpoint in #1253, with the complete guest-to-account flow delivered across the stack.

PR #1286, opened 2026-09-02
Fix typo in confetti submission label

azzy9:
Simple typo fix that was bugging me for "Confetti on successful submission" label.


JhumanJ (approved):
Good catch haha, thanks for the PR!

